PR #83, opened 2021-10-18
Centreon tests improvement

bouda1:
centreon-tests has several tests improved/fixed.


bouda1:
We give up with jest.

PR #99, opened 2021-11-19
fix(rrd): add sql query to check metrics to delete

rem31:


centreon-broker/storage/src/conflict_manager_storage.cc
```diff
@@ -568,6 +568,16 @@ void conflict_manager::_check_deleted_index() {
         _metric_cache.erase({res.value_as_u64(0), res.value_as_str(2)});
         _index_cache.erase({res.value_as_u32(3), res.value_as_u32(4)});
       }
+      _mysql.run_query_and_get_result(
+          "SELECT m.metric_id, m.metric_name FROM metrics m WHERE "
+          "m.to_delete=1",
```
bouda1, line 573: SELECT metric_id, metric_name FROM metrics WHERE to_delete=1

bouda1 (changes requested):

PR #109, opened 2021-11-30
sql connection uptime in broker stats

Maaown:


bouda1:
The way stats are get does not work because of a bad choice in the architecture. I close this PR and a new one should arrive quickly.

PR #159, opened 2022-02-25
fix(flapping): add check flapping service on soft state

rem31:


centreon-engine/src/service.cc
```diff
@@ -2040,12 +2040,6 @@ void service::check_for_flapping(bool update,
   log_v2::checks()->debug("Checking service '{}' on host '{}' for flapping...",
                           _description, _hostname);
 
-  /* if this is a soft service state and not a soft recovery, don't record this
-   * in the history */
-  /* only hard states and soft recoveries get recorded for flap detection */
-  if (get_state_type() == soft && _current_state != service::state_ok)
-    return;
-
```
bouda1, line 2048: This fix looks good.
Anyway, I would like to see a unit test that validates this new behavior.
We have the file `centreon-engine/tests/notifications/service_flapping_notification.cc` that contains many tests on flapping.
If you take a look at one test, say `TEST_F(ServiceFlappingNotification, CheckFlappingWithHostDown)`,
you will see 3 steps:
* We empty the flap stack by setting the service to OK many times.
* We set the service to CRITICAL then to OK, this is repeated 8 times. We use a trick for this test to work, the `max attempts` is fixed to 1. With your fix, I think this trick could be removed.

So you can add a new test by copying this previous test and setting `max_attempts` to 3 ; and see if the flap detection still works. And others tests should still work too!

bouda1 (changes requested):

PR #160, opened 2022-02-25
enh(tests): Cops 824 cicd create pipeline centreon test on centreon collect

zguennoune02:
Adding centreon tests in our pipeline and run robot framework tests every day at midnight


bouda1 (commented):
Have you tried to compile collect after these changes?
We have a Python script that reads the sql file to generate values. It works with some regexp and could fail with your changes. So that's why I ask the question.
Otherwise, all looks good!

bouda1 (approved):

bouda1 (approved):

PR #180, opened 2022-03-28
enh(engine/tags): we can add a tags to a service and host now.

rem31:
REFS: MON-12287


broker/unified_sql/src/stream_sql.cc
```diff
@@ -2858,8 +2921,44 @@ void stream::_process_tag(const std::shared_ptr<io::data>& d) {
   if (tg.action() != Tag_Action_DELETE) {
     *st << *s;
     int32_t conn = special_conn::tag % _mysql.connections_count();
-    _mysql.run_statement(*st, database::mysql_error::store_tag, false, conn);
+    std::promise<uint64_t> p;
```
bouda1, line 2924: Because of last modifications in the DB, this way to update the db won't work. Please look at my changes in the severity code and you should be able to adapt this code.

broker/unified_sql/src/stream_sql.cc
```diff
@@ -2858,8 +2921,44 @@ void stream::_process_tag(const std::shared_ptr<io::data>& d) {
...
+        }
+      } catch (const std::exception& e) {
+      }
+      if (tag_id == 0)
+        throw exceptions::msg_fmt("Could not fetch tag_id of tag ({}, {}): {}",
+                                  tg.id(), tg.type(), e.what());
+    }
+    log_v2::sql()->debug("juste avant pb insert cache ({}, {}, tag_id {})",
+                         tg.id(), tg.type(), tag_id);
+    tags_cache.emplace(std::make_pair(tg.id(), tg.type()), tag_id);
```
bouda1, line 2961: no need to update tags_cache each time, just when tag_id is new.

engine/inc/com/centreon/engine/configuration/host.hh
```diff
@@ -113,6 +113,8 @@ class host : public object {
   unsigned int stalking_options() const throw();
   std::string const& statusmap_image() const throw();
   std::string const& timezone() const throw();
+  void tags(const std::string& value);
+  const std::list<uint64_t>& tags() const noexcept;
```
bouda1, line 117: I have a doubt here, don't you also need the type of the tag ?

engine/inc/com/centreon/engine/configuration/host.hh
```diff
@@ -218,6 +221,7 @@ class host : public object {
   std::string _statusmap_image;
   opt<std::string> _timezone;
   std::string _vrml_image;
+  std::list<uint64_t> _tags;
```
bouda1, line 224: how to deal with several types of tag if you just have an id...

engine/inc/com/centreon/engine/configuration/service.hh
```diff
@@ -123,6 +123,8 @@ class service : public object {
   unsigned short stalking_options() const throw();
   void timezone(std::string const& time_zone);
   std::string const& timezone() const throw();
+  void tags(const std::string& value);
+  const std::list<uint64_t>& tags() const noexcept;
```
bouda1, line 127: same issue

engine/inc/com/centreon/engine/configuration/service.hh
```diff
@@ -222,6 +225,7 @@ class service : public object {
   static std::unordered_map<std::string, setter_func> const _setters;
   opt<unsigned short> _stalking_options;
   opt<std::string> _timezone;
+  std::list<uint64_t> _tags;
```
bouda1, line 228: idem

engine/inc/com/centreon/engine/tag.hh
```diff
@@ -31,7 +31,8 @@ class tag;
 }
 
 using tag_map =
-    absl::flat_hash_map<uint32_t, std::shared_ptr<com::centreon::engine::tag>>;
+    absl::flat_hash_map<std::pair<uint64_t, uint32_t>,
+                        std::shared_ptr<com::centreon::engine::tag>>;
```
bouda1, line 35: Why using a map?
Do you need to look for a specific tag from its id/type? I don't think so.
So, a std::list should be more appropriate.

engine/src/configuration/applier/host.cc
```diff
@@ -134,6 +134,27 @@ void applier::host::add_object(configuration::host const& obj) {
     }
   }
 
+  // add tags
+  for (auto tag : obj.tags()) {
+    tag_map::iterator it_tag_group{engine::tag::tags.find(
+        std::make_pair(tag, engine::tag::tagtype::hostgroup))};
+    tag_map::iterator it_tag_category{engine::tag::tags.find(
+        std::make_pair(tag, engine::tag::tagtype::hostcategory))};
```
bouda1, line 142: this code is not efficient

bouda1: A list would be better

engine/inc/com/centreon/engine/service.hh
```diff
@@ -192,6 +193,8 @@ class service : public notifier {
   host const* get_host_ptr() const;
   host* get_host_ptr();
   bool get_host_problem_at_last_check() const;
+  tag_map& mut_tags() noexcept;
+  const tag_map& tags() const noexcept;
```
bouda1, line 197: service and host inherits from checkable.
So, tags would be better in checkable, then you won't write the same code twice...

bouda1 (changes requested):

engine/src/configuration/host.cc
```diff
@@ -1712,6 +1729,54 @@ bool host::_set_timezone(std::string const& value) {
...
+ *  Set host tags.
+ *
+ *  @param[in] value  The new tags.
+ *
+ *  @return True.
+ */
+bool host::_set_category_tags(const std::string& value) {
+  bool ret;
+  std::list<absl::string_view> tags{absl::StrSplit(value, ',')};
+  for (auto tag : _tags) {
```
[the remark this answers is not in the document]
bouda1: Would std::erase_if or std::remove_if be better?

engine/src/configuration/host.cc
```diff
@@ -1712,6 +1729,54 @@ bool host::_set_timezone(std::string const& value) {
...
+    if (tag.second == tag::hostcategory) {
+      _tags.erase(tag);
+    }
+  }
+
+  for (auto tag : tags) {
```
bouda1, line 1748: auto&

engine/src/configuration/host.cc
```diff
@@ -1712,6 +1729,54 @@ bool host::_set_timezone(std::string const& value) {
...
+    int64_t id;
+    ret = SimpleAtoi(tag, &id);
```
bouda1, line 1750: if tag is not an integer, what happens?

engine/src/configuration/host.cc
```diff
@@ -1712,6 +1729,54 @@ bool host::_set_timezone(std::string const& value) {
...
+bool host::_set_group_tags(const std::string& value) {
+  bool ret;
+  std::list<absl::string_view> tags{absl::StrSplit(value, ',')};
+  for (auto tag : _tags) {
+    if (tag.second == tag::hostgroup) {
+      _tags.erase(tag);
+    }
+  }
+
+  for (auto tag : tags) {
```
bouda1, line 1772: auto&

engine/src/configuration/host.cc
```diff
@@ -1712,6 +1729,54 @@ bool host::_set_timezone(std::string const& value) {
...
+ *  @return True.
+ */
+bool host::_set_group_tags(const std::string& value) {
+  bool ret;
+  std::list<absl::string_view> tags{absl::StrSplit(value, ',')};
+  for (auto tag : _tags) {
+    if (tag.second == tag::hostgroup) {
+      _tags.erase(tag);
+    }
+  }
```
bouda1, line 1770: better with iterators and std::algorithms

engine/src/configuration/service.cc
```diff
@@ -2110,6 +2134,54 @@ bool service::_set_timezone(std::string const& value) {
...
+  }
+
+  for (auto tag : tags) {
+    int64_t id;
+    ret = SimpleAtoi(tag, &id);
+    _tags.emplace(id, tag::servicecategory);
+  }
+  return ret;
+}
+
```
bouda1, line 2160: idem

engine/src/configuration/service.cc
```diff
@@ -2110,6 +2134,54 @@ bool service::_set_timezone(std::string const& value) {
...
+    }
+  }
+
+  for (auto tag : tags) {
+    int64_t id;
+    ret = SimpleAtoi(tag, &id);
+    _tags.emplace(id, tag::servicegroup);
+  }
+  return ret;
+}
```
bouda1, line 2183: idem

engine/src/configuration/tag.cc
```diff
@@ -41,15 +41,15 @@ const absl::flat_hash_map<std::string, tag::setter_func> tag::_setters{
  * @param key The unique id corresponding to this tag.
  */
 tag::tag(const key_type& key)
-    : object(object::tag), _id{key}, _type{tag::none} {}
+    : object(object::tag), _key{key}, _type{tag::none} {}
```
bouda1, line 44: I thought the key holds id and type.
In that case, why also having a type?

tests/resources/Engine.py
```diff
@@ -560,3 +570,27 @@ def remove_severities_from_services():
     ff = open("{}/config0/services.cfg".format(CONF_DIR), "w")
     ff.writelines(out)
     ff.close()
+
+def add_tags_to_host(type:str, tag_id:str, hst_lst):
```
bouda1, line 574: hosts would be better

bouda1 (changes requested):

broker/storage/src/conflict_manager_sql.cc
```diff
@@ -42,7 +42,11 @@ using namespace com::centreon::broker::storage;
 void conflict_manager::_clean_tables(uint32_t instance_id) {
   /* Database version. */
 
-  int32_t conn = _mysql.choose_connection_by_instance(instance_id);
+  int32_t conn = special_conn::tag % _mysql.connections_count();
+  _mysql.run_query("DELETE FROM resources_tags",
+                   database::mysql_error::clean_resources_tags, false, conn);
+
+  conn = _mysql.choose_connection_by_instance(instance_id);
```
bouda1, line 49: I have a big doubt on this query

bouda1 (approved):

PR #197, opened 2022-04-20
Mon 12190 resources tags severities

bouda1:
many bugs fixed in insertions into resources.

REFS: MON-12190


broker/core/src/file/directory_watcher.cc
```diff
@@ -141,11 +142,11 @@ std::vector<directory_event> directory_watcher::get_events() {
   }
   log_v2::core()->debug("file: directory watcher getting events of size {}",
                         buf_size);
-  char* buf = new char[buf_size];
+  char* buf = (char*)alloca(buf_size);
   int len = ::read(_inotify_instance_id, buf, buf_size);
   if (len == -1) {
     int err = errno;
-    delete[] buf;
+    // delete[] buf;
```
rem31, line 149: is required?

rem31 (changes requested):

PR #201, opened 2022-04-22
enh(tests): add stability for rrd tests

rem31:
REFS: MON-12158

### DIFF
--- a/broker/unified_sql/src/stream_sql.cc
+++ b/broker/unified_sql/src/stream_sql.cc
@@ -58,8 +58,10 @@ void stream::_clean_tables(uint32_t instance_id) {
   }
 
   conn = _mysql.choose_connection_by_instance(instance_id);
-  _mysql.run_query(fmt::format(
-      "UPDATE resources SET enabled=0 WHERE poller_id={}", instance_id), database::mysql_error::clean_resources, false, conn);
+  _mysql.run_query(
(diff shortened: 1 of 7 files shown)


broker/unified_sql/src/stream_sql.cc
```diff
@@ -2650,7 +2652,7 @@ void stream::_check_and_update_index_cache(const Service& ss) {
     _index_data_insert.bind_value_as_i32(2, ss.service_id());
     _index_data_insert.bind_value_as_str(3, sv);
     _index_data_insert.bind_value_as_str(4, "0");
-    _index_data_insert.bind_value_as_bool(5, special);
+    _index_data_insert.bind_value_as_bool(5, special ? "1" : "0");
```
bouda1, line 2655: this method has the following prototype:
void mysql_stmt::bind_value_as_bool(int range, bool value);
As you can see the second argument is a boolean and not a char*.
In your case, this is cast to boolean and as "1" or "0" is not a nullptr, they are both considered as true.

bouda1 (changes requested):

broker/unified_sql/src/stream_sql.cc
bouda1: So I think you introduced a bug here.

bouda1 (approved):

PR #206, opened 2022-04-26
Mon 12158 enforce tests dbo2

bouda1:
* Graphes are back in passive services
* Segfault fixed in engine during extinction
* Various fixes in centreon-tests

REFS: MON-12158 MON-13163


broker/rrd/src/creator.cc
```diff
@@ -156,7 +156,7 @@ void creator::create(std::string const& filename,
  */
 void creator::_duplicate(std::string const& filename, fd_info const& in_fd) {
   // Remove previous file.
-  remove(filename.c_str());
+  //remove(filename.c_str());
```
rem31: it's ok ?

rem31 (approved):

PR #249, opened 2022-05-31
add clib diretory to load from ldconfig

lgcosta:


zguennoune02 (approved):

ci/debian/centreon-clib.postinst
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 if [ "$1" = "configure" ] ; then
+    echo "/usr/lib64" >> /etc/ld.so.conf.d/centreon.conf
```
bouda1, line 4: This is not standard at all on Debian.
Is it possible to install the clib in the /usr/lib directory for Debian, this would far better!

lgcosta: @bouda1 can we configure this in the build (`./configure`) ? or do I have to force it in the installation (move file or create symbolic link to `/usr/lib`) ?

PR #253, opened 2022-06-02
MON-12283 add AddHold and some tests and trace

jean-christophe81:


broker/grpc/src/client.cc
```diff
@@ -99,6 +99,7 @@ client::~client() {
 void client::start_read(event_ptr& to_read, bool first_read) {
   StartRead(to_read.get());
   if (first_read) {
+    AddHold();
```
bouda1, line 102: In the documentation, they also speak about RemoveHold(), but this is not present in your code...

jean-christophe81: added in case of communication failure

bouda1 (approved):

PR #256, opened 2022-06-02
enh(sonar): replace atoi by abseil::SimpleAtoi()

denliA:
replaced atoi by SimpleAtoi() and throw exception if entry != positive integer

fixed getopt_long() call, 't' doesn't exist so replaced by 's'

Debug seems unused

getopt_long reads only one option and ignores the rest, maybe put it inside a while loop until EOF


bouda1 (changes requested):
I would like to see a centreon-tests that validates this behavior.

bouda1 (approved):

PR #361, opened 2022-09-13
fix(broker/unified_sql): cbd no more blocked when a bad row is encoun…

bouda1:
…tered in db

## Description

cbd is more strict when it reads the db. It can throw exception if unexpected values are found. This breaks several centreon platforms. So we changed a little this behaviour to raise errors but not break cbd.

REFS: MON-14951

## Type of change

### DIFF
--- a/broker/unified_sql/src/stream.cc
+++ b/broker/unified_sql/src/stream.cc
@@ -163,9 +163,16 @@ void stream::_load_deleted_instances() {
   _mysql.run_query_and_get_result(query, std::move(promise));
   try {
     mysql_result res(future.get());
-    while (_mysql.fetch_row(res))
-      _cache_deleted_instance_id.insert(res.value_as_u32(0));
-  } catch (std::exception const& e) {
+    while (_mysql.fetch_row(res)) {
+      int32_t instance_id = res.value_as_i32(0);
+      if (instance_id <= 0)
+        log_v2::sql()->error(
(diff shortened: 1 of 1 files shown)


broker/unified_sql/src/stream.cc
```diff
@@ -639,12 +695,28 @@ void stream::remove_graphs(const std::shared_ptr<io::data>& d) {
         std::lock_guard<misc::shared_mutex> lock(_metric_cache_m);
         while (ms.fetch_row(res)) {
           indexes_to_delete.insert(res.value_as_u64(0));
-          uint64_t mid = res.value_as_u64(1);
-          if (mid)
+          int64_t mid = res.value_as_i32(1);
```
jean-christophe81, line 698: as_i64

broker/unified_sql/src/stream.cc
```diff
@@ -661,8 +733,15 @@ void stream::remove_graphs(const std::shared_ptr<io::data>& d) {
 
         std::lock_guard<misc::shared_mutex> lock(_metric_cache_m);
         while (ms.fetch_row(res)) {
-          metrics_to_delete.insert(res.value_as_u64(1));
-          _metric_cache.erase({res.value_as_u64(0), res.value_as_str(2)});
+          int32_t metric_id = res.value_as_i32(1);
```
jean-christophe81, line 736: as_i64

jean-christophe81 (approved):

PR #399, opened 2022-10-11
Mon 12982 downtime by

bouda1:
## Description

Downtimes are stored with hosts and services id now.

REFS: MON-12982

## Type of change

- [ ] Patch fixing an issue (non-breaking change)


engine/src/commands/commands.cc
```diff
@@ -855,132 +855,129 @@ int cmd_remove_acknowledgement(int cmd, char* args) {
...
   time_t end_time{0};
   bool fixed{false};
   uint64_t triggered_by{0};
   unsigned long duration{0};
-  char* author{nullptr};
-  char* comment_data{nullptr};
   uint64_t downtime_id{0};
   servicegroup_map::const_iterator sg_it;
 
+  std::vector<absl::string_view> a{absl::StrSplit(args, ';')};
```
jean-christophe81, line 869: Is auto more efficient to avoid conversion to vector?

engine/src/commands/commands.cc
```diff
@@ -1146,57 +1146,75 @@ int cmd_delete_downtime(int cmd, char* args) {
  *  @param[in] args  Command arguments.
  */
 int cmd_delete_downtime_full(int cmd, char* args) {
-  char* temp_ptr(nullptr);
+  log_v2::functions()->trace("cmd_delete_downtime_full() args = {}", args);
   downtime_finder::criteria_set criterias;
 
+  std::vector<absl::string_view> a{absl::StrSplit(args, ';')};
```
jean-christophe81, line 1152: idem

engine/src/comment.cc
```diff
@@ -114,16 +119,20 @@ void comment::delete_host_comments(uint64_t host_id) {
           it->second->get_comment_data().c_str(), it->second->get_persistent(),
           it->second->get_source(), it->second->get_expires(),
           it->second->get_expire_time(), it->first);
-      it = comments.erase(it);
+      comments.erase(it++);
```
jean-christophe81, line 122: it = comments.erase(it)

engine/src/comment.cc
```diff
@@ -134,19 +143,22 @@ void comment::delete_service_comments(uint64_t host_id, uint64_t service_id) {
...
```
jean-christophe81, line 146: idem

engine/src/comment.cc
```diff
@@ -157,21 +169,23 @@ void comment::delete_host_acknowledgement_comments(engine::host* hst) {
...
```
jean-christophe81, line 172: idem

engine/src/comment.cc
bouda1: this the recommanded way by absl. Their method is a void method.

bouda1: this would not compile

engine/src/comment.cc
bouda1: this would not compile

engine/src/comment.cc
bouda1: would not compile

engine/src/comment.cc
```diff
@@ -183,9 +197,9 @@ void comment::delete_service_acknowledgement_comments(::service* svc) {
...
```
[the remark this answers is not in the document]
bouda1: would not compile

PR #401, opened 2022-10-12
Mon 15408 log entry

bouda1:
## Description

log entry migrated to protobuf.

REFS: MON-15408

## Type of change

### DIFF
--- a/broker/neb/src/set_log_data.cc
+++ b/broker/neb/src/set_log_data.cc
@@ -17,6 +17,8 @@
 */
 
 #include "com/centreon/broker/neb/set_log_data.hh"
+#include <absl/strings/str_split.h>
+#include "com/centreon/broker/log_v2.hh"
 #include "com/centreon/broker/neb/internal.hh"
 #include "com/centreon/broker/neb/log_entry.hh"
 #include "com/centreon/engine/host.hh"
@@ -47,6 +49,28 @@ static char* log_extract(char** lasts) {
 }
 
 /**
+ * @brief Get the id of a log status.
+ *
+ * @param status A string corresponding to the state of the resource.
+ *
+ * @return A status code.
+ */
+static int status_id(const absl::string_view& status) {
+  int retval;
+  if (status == "DOWN" || status == "WARNING")
+    retval = 1;
+  else if (status == "UNREACHABLE" || status == "CRITICAL")
+    retval = 2;
+  else if (status == "UNKNOWN")
+    retval = 3;
+  else if (status == "PENDING")
+    retval = 4;
+  else
+    retval = 0;
+  return retval;
+}
+
+/**
  *  Get the id of a log status.
  */
 static int status_id(char const* status) {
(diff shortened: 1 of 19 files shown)


broker/neb/src/set_log_data.cc
```diff
@@ -46,6 +48,28 @@ static char* log_extract(char** lasts) {
...
+  else if (status == "PENDING")
+    retval = 4;
+  else
+    retval = 0;
+  return retval;
+}
+
 /**
  *  Get the id of a log status.
  */
```
jean-christophe81, line 75: why not use the new version of status_id since string_view(const char *) exists?

broker/neb/src/set_log_data.cc
```diff
@@ -64,6 +88,21 @@ static int status_id(char const* status) {
...
+    absl::string_view nstatus = status.substr(pos_start, pos_end - pos_start);
+    retval = status_id(nstatus);
+  } else
+    retval = status_id(status);
+  return retval;
+}
+
 /**
  *  Get the notification status of a log.
  */
```
jean-christophe81, line 108: idem

broker/neb/src/set_log_data.cc
```diff
@@ -224,3 +275,628 @@ void neb::set_log_data(neb::log_entry& le, char const* log_data) {
...
+  auto ait = args.begin();
+
+  if (typ == "SERVICE ALERT") {
+    le_obj.set_msg_type(LogEntry_MsgType_SERVICE_ALERT);
+    if (ait == args.end()) {
+      log_v2::neb()->error("Missing host name in log message '{}'",
+                           le_obj.output());
+      return false;
+    }
+    le_obj.set_host_name(std::string(ait->data(), ait->size()));
```
jean-christophe81, line 305: you can avoid std::string with le_obj.set_host_name(ait->data(), ait->size());

broker/neb/src/set_log_data.cc
```diff
@@ -224,3 +275,628 @@ void neb::set_log_data(neb::log_entry& le, char const* log_data) {
...
+    ++ait;
+
+    if (ait == args.end()) {
+      log_v2::neb()->error("Missing service description in log message '{}'",
+                           le_obj.output());
+      return false;
+    }
+    le_obj.set_service_description(std::string(ait->data(), ait->size()));
```
jean-christophe81, line 313: idem

broker/neb/src/set_log_data.cc
```diff
@@ -224,3 +275,628 @@ void neb::set_log_data(neb::log_entry& le, char const* log_data) {
...
+    ++ait;
+
+    if (ait == args.end()) {
```
jean-christophe81, line 316: replace all these if (ait == args.end()){
....
} by a macro?

broker/neb/src/set_log_data.cc
bouda1: I was wondering the same question... So, let's go for it

broker/neb/src/set_log_data.cc
bouda1: agree

broker/neb/src/set_log_data.cc
bouda1: :+1:

broker/neb/src/set_log_data.cc
bouda1: I was lazy, the original code is very long to read, and I just made replacement in it.
The new one is faster (no more copy), I will try try to make something more readable.

jean-christophe81 (approved):

PR #437, opened 2022-11-09
feat(chore): update .gitignore and remove deprecated scripts

sc979:
## Description

Update gitignore and remove deprecated scripts

**Fixes** # (none)

## Type of change

## Target serie

- [ ] 21.04.x
- [x] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x (master)


.gitignore
```diff
@@ -1,13 +1,16 @@
...
 *.so
 *.swp
-.idea
-.vscode
+
 CMakeCache.txt
 CMakeFiles
 CTest*.cmake
 Makefile
+.conan
```
bouda1, line 13: Why this .conan line?
I know only two cases for this folder to be created:
* root case: it is created in /root/.conan
* user case: it is created in /home/USER/.conan

Do you hace centreon-collect at the same place?

sc979: It came from the backport. Originaly it has been added here : https://github.com/centreon/centreon-collect-gha/blob/develop/.gitignore#L13

Do not hesitate to suggest directives removal ;)

PR #475, opened 2022-12-13
Next 21.10 release #18909#

paloth:
Centreon 21.10 next release #18909#


bdauria:
@chgautier @bouda1 what's the status for this release?

chgautier:
There's a coredump issue that was found just before the holidays that needs to be fixed. @bdauria

PR #495, opened 2023-01-04
fix(engine/ack): sometimes acknowledgement_type and problem_has_been_…

bouda1:
…acknowledged were desynchronized

## Description

This patch is a fix in notifier. The idea is to merge the boolean _problem_has_been_acknowledged and the enum _acknowledgement_type (which can be NONE, NORMAL, STICKY).

The NONE type is equivalent to the problem not acknowledged.

REFS: MON-16401

## Type of change

[MON-16401]: https://centreon.atlassian.net/browse/MON-16401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


engine/inc/com/centreon/engine/notifier.hh
```diff
@@ -45,6 +45,12 @@ class notifier : public checkable {
     cat_custom,
   };
 
+  enum acknowledgement_type {
+    ACK_NONE = 0,
+    ACK_NORMAL = 1,
+    ACK_STICKY = 2,
+  };
+
```
jean-christophe81, line 53: why not use ServiceStatus_AckType enum?

bbdo/host.proto
```diff
@@ -31,7 +31,7 @@ message Host {
     STICKY = 2;
   }
```
jean-christophe81, line 32: is it better to have this common enum defined outside Host and Service and used by noth?

jean-christophe81 (approved):

PR #539, opened 2023-01-27
fix(ci): handle delivery for every enterprise linux

kduret:
## Description

handle delivery for every enterprise linux (not only el7 and el8)

**Fixes** MON-16456

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [x] 23.04.x (master)

Check github actions

### DIFF
--- a/.github/workflows/centreon-collect.yml
+++ b/.github/workflows/centreon-collect.yml
@@ -132,12 +132,6 @@ jobs:
           username: ${{ secrets.REPOS_USERNAME }}
           password: ${{ secrets.REPOS_PASSWORD }}
 
-      - name: Upload package artifacts
-        uses: actions/upload-artifact@v3
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/centreon-collect.yml
```diff
@@ -132,12 +132,6 @@ jobs:
           username: ${{ secrets.REPOS_USERNAME }}
           password: ${{ secrets.REPOS_PASSWORD }}
 
-      - name: Upload package artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          name: packages-rpm-${{ matrix.distrib }}
-          path: ./*.rpm
-
```
bouda1, line 140: Do we still have an artifact and this would be a second definition?

bouda1 (approved):

PR #665, opened 2023-03-27
Next 22.04 release #19056

paloth:
Centreon 22.04 next release #19056#


chgautier:
@bouda1 @kduret I disabled the nightly workflow by only letting it trigger on develop. We can reenable it afterwards if it feasible for dev-22.04.x.

PR #723, opened 2023-05-02
Mon 18447 bulk on data bin 22.04

bouda1:
## Description

* to insert into data_bin, we use bulk statements now when available.
* New grpc command to observe mysql connections.

REFS: MON-18896


bouda1:
This branch was rewritten.

PR #880, opened 2023-09-07
MON-21347 Sync Centreon next 23.04 (Jira release #21347#)

tuntoja:
REF #MON-21347


tuntoja:
Merged considering test EBBPS2 is currently known as being faulty (the test is, not the application). Discussed with @jean-christophe81

PR #921, opened 2023-10-12
enh(SAST): Move scan analysis in robot job

sc979:
## Description

*relates to* : #884

Improve analysis consistency and trigger nightly analysis only on develop maintenance branch
Avoid to fail pipelines after successive merge on develop

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x (master)


.github/workflows/veracode-analysis.yml
```diff
@@ -73,10 +69,15 @@ jobs:
...
+          echo "[DEBUG] - Find compiled files"
+          find ./ -name "*.so"
+          echo "[DEBUG] - Build size"
+          du -sh ./lib/*.so | sort -rh
+
       - name: Binary preparation
         run: |
           echo "[INFO] - Keeping only compiled files"
-          find ./build -type f -not \( -name "*.so" -o -name "*.h" -o -name "*.cc" -o -name "*.hpp" \) -delete
+          find ./build -type f -not \( -name "*.so" \) -delete
```
bouda1, line 80: Maybe I'm going to say something totally stupid. You're warned !
Here you remove in build files that are not *.so
But we also have binaries in build/bin...
Don't you check them?
And in build/lib, we have *.a files that are static libraries, so equivalent to *.so but static instead of dynamic.

I have the feeling you don't check everything...

.veracode-exclusions
```diff
@@ -35,3 +35,4 @@ engine/src/simumod/coder.pl
 grpc/centreon-rpc-client.py
 tests
 packaging
+/root/.conan/
```
bouda1, line 38: protobuf generates source code files from our *.proto files of the form *.pb.cc and *.pb.h
I think these files should also been excluded.

sc979: currently the script cannot use wildcards to exclude files.
if more exclusions are needed, you will be able to add them to this file, after the merge or right now 👍

.github/workflows/veracode-analysis.yml
sc979: Indeed, the first objective of this task was to parallelize the analysis.
In veracode I could only find .so file analysed. But we can try to add others.
We can also trick the engine to force the analysis by renaming the static libraries from .a to .so 🤔

sc979: to add : 
/build/bin (at least cbd)

sc979: added to binary to analyze

kduret (approved):

tuntoja (approved):

PR #1001, opened 2023-12-12
enh(ci): PoC sccache

kduret:
## Description

enh(ci): PoC sccache

**Fixes** MON-32903

## Type of change

- [ ] Patch fixing an issue (non-breaking change)
- [x] New functionality (non-breaking change)

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [x] 24.04.x (master)

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -5,16 +5,17 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  pull_request:
-    branches:
-      - develop
-      - dev-[2-9][0-9].[0-9][0-9].x
-      - master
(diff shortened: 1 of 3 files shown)


kduret (commented):
to validate sccache is working as expected, you should test following cases :
- make a small change in this [file](https://github.com/centreon/centreon-collect/blob/develop/broker/neb/inc/com/centreon/broker/neb/acknowledgement.hh) and check cache is still used
- duplicate this pull request and check cache is still used

A retention rule should probably be apply on S3 bucket to infinite size incrementation

PR #1041, opened 2024-01-05
Mon 33636 ba status not updated develop

bouda1:
## Description

If a worst BA has one child that is also a worst BA, when the child changes to a state, the parent should also change to that state. It wasn't always the case.
Several reasons:
* Since the previous change in bam it is the child that checks if it changes and should notify its parent. But there was still a piece of code in the parent that also made the check. So we could have some changes not detected.
* The change of a kpi in  a BA was done by two function, unapply_impact() and apply_impact(), but only apply_impact() returned a boolean to tell if there was a change. Now we have a new method apply_changes() that returns a boolean and makes directly all the needed changes.

REFS: MON-33636

### DIFF
--- a/broker/bam/src/ba_best.cc
+++ b/broker/bam/src/ba_best.cc
@@ -1,5 +1,5 @@
 /**
- * Copyright 2022 Centreon
+ * Copyright 2022-2024 Centreon
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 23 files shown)


thearchit3ct (approved):

broker/bam/src/ba_best.cc
```diff
@@ -96,29 +97,54 @@ state ba_best::get_state_soft() const {
...
+ */
+bool ba_best::_apply_changes(kpi* child,
+                             const impact_values& new_hard_impact,
+                             const impact_values& new_soft_impact,
+                             bool in_downtime) {
+  state previous_state = _computed_hard_state;
+
+  _computed_soft_state = _computed_hard_state = state_critical;
+
+  // We recompute all impacts...
```
jean-christophe81, line 137: why recompute all impacts on change of only one kpi?

broker/bam/src/ba_ratio_number.cc
```diff
@@ -123,14 +120,37 @@ void ba_ratio_number::_unapply_impact(kpi* kpi_ptr,
...
+ */
+bool ba_ratio_number::_apply_changes(kpi* child,
+                                     const impact_values& new_hard_impact,
+                                     const impact_values& new_soft_impact,
+                                     bool in_downtime) {
+  double previous_level = _level_hard;
+  _level_soft = 0.;
+  _level_hard = 0.;
+
+  // We recompute all impact, except the one to unapply...
```
jean-christophe81, line 142: idem

broker/bam/src/ba_ratio_percent.cc
```diff
@@ -125,14 +122,37 @@ void ba_ratio_percent::_unapply_impact(kpi* kpi_ptr,
...
+ */
+bool ba_ratio_percent::_apply_changes(kpi* child,
+                                      const impact_values& new_hard_impact,
+                                      const impact_values& new_soft_impact,
+                                      bool in_downtime) {
+  double previous_level = _level_hard;
+  _level_soft = 0.;
+  _level_hard = 0.;
+
+  // We recompute all impact, except the one to unapply...
```
jean-christophe81, line 144: idem

broker/bam/src/ba_worst.cc
```diff
@@ -96,29 +97,59 @@ state ba_worst::get_state_soft() const {
...
+ */
+bool ba_worst::_apply_changes(kpi* child,
+                              const impact_values& new_hard_impact,
+                              const impact_values& new_soft_impact,
+                              bool child_in_downtime) {
+  state previous_state = _computed_hard_state;
+
+  _computed_soft_state = _computed_hard_state = state_ok;
+
+  // We recompute all impacts...
```
jean-christophe81, line 141: idem

broker/bam/src/ba_best.cc
bouda1: Not all, just the one of the BA. It is of type BEST. So we have to all the children to get the best one, haven't we?

broker/bam/src/ba_ratio_number.cc
bouda1: Same answer

jean-christophe81 (approved):

PR #1159, opened 2024-02-23
Mon 35113 peer list 22.10

bouda1:
## Description

Peers list in broker stats are correctly updated now.

REFS: MON-35113


bouda1:
REFS: MON-35114

PR #1166, opened 2024-02-27
Mon 36693 new robot tests on downtimes 23.10

bouda1:
REFS: MON-36693


kduret (approved):
I guess my review is needed because the base branch has been changed

PR #1399, opened 2024-06-05
enh(engine/configuration): use of misc::string removed from configura…

bouda1:
…tion

REFS: MON-85674

### DIFF
--- a/engine/src/configuration/parser.cc
+++ b/engine/src/configuration/parser.cc
@@ -365,7 +365,7 @@ void parser::_parse_object_definitions(std::string const& path) {
   std::string input;
   while (get_next_line(stream, input, _current_line)) {
     // Multi-line.
-    while (!input.empty() && '\\' == input[input.size() - 1]) {
+    while ('\\' == input[input.size() - 1]) {
       input.resize(input.size() - 1);
       std::string addendum;
       if (!get_next_line(stream, addendum, _current_line))
--- a/engine/src/configuration/parser.cc
+++ b/engine/src/configuration/parser.cc
@@ -365,7 +365,7 @@ void parser::_parse_object_definitions(std::string const& path) {
   std::string input;
   while (get_next_line(stream, input, _current_line)) {
     // Multi-line.
-    while ('\\' == input[input.size() - 1]) {
+    while (!input.empty() && '\\' == input[input.size() - 1]) {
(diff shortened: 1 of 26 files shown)


engine/src/configuration/parser.cc
```diff
@@ -334,12 +363,12 @@ void parser::_parse_object_definitions(std::string const& path) {
   bool parse_object = false;
   object_ptr obj;
   std::string input;
-  while (string::get_next_line(stream, input, _current_line)) {
+  while (get_next_line(stream, input, _current_line)) {
     // Multi-line.
-    while ('\\' == input[input.size() - 1]) {
+    while (!input.empty() && '\\' == input[input.size() - 1]) {
```
jean-christophe81, line 368: input is never empty if get_next_line does not return false

engine/src/configuration/state.cc
```diff
@@ -38,13 +37,37 @@ struct setter : public setter_base {
...
+      } else if constexpr (std::is_same_v<U, double>) {
+        if (!absl::SimpleAtod(value, &val))
+          return false;
+        (obj.*ptr)(val);
+      } else {
+        static_assert(std::is_integral_v<U> || std::is_floating_point_v<U> ||
+                      std::is_same_v<U, bool>);
+      }
+    } catch (const std::exception& e) {
+      config_logger->error(e.what());
```
jean-christophe81, line 70: a better log like config_logger->error("fail to parse {}={}: {}", setter_base::_field_name, value, e.what());

engine/src/configuration/timeperiod.cc
```diff
@@ -264,19 +262,23 @@ days_array const& timeperiod::timeranges() const {
  */
 bool timeperiod::_build_timeranges(std::string const& line,
                                    timerange_list& timeranges) {
-  list_string timeranges_str;
-  string::split(line, timeranges_str, ',');
-  for (list_string::const_iterator it(timeranges_str.begin()),
-       end(timeranges_str.end());
-       it != end; ++it) {
-    std::size_t pos(it->find('-'));
+  size_t pos = line.find_first_of(';');
```
jean-christophe81, line 265: no comment allowed now in cfg files

jean-christophe81 (approved):

PR #1439, opened 2024-06-13
Mon 34072 dependencies escalations reviewed

bouda1:
## Description

Hostescalations, serviceescalations, hostdependencies and servicedependencies had a uuid as key when they were stored as configuration. It is no more used and replaced by a Hash built from their attributes. So now, we can find this hash directly from the object, this simplifies the code and remove a dependency to the engine code.

REFS: MON-34072

### DIFF
--- a/engine/inc/com/centreon/engine/hostdependency.hh
+++ b/engine/inc/com/centreon/engine/hostdependency.hh
@@ -1,22 +1,21 @@
-/*
-** Copyright 2011-2019 Centreon
-**
-** This file is part of Centreon Engine.
-**
-** Centreon Engine is free software: you can redistribute it and/or
-** modify it under the terms of the GNU General Public License version 2
-** as published by the Free Software Foundation.
-**
-** Centreon Engine is distributed in the hope that it will be useful,
-** but WITHOUT ANY WARRANTY; without even the implied warranty of
-** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
-** General Public License for more details.
-**
-** You should have received a copy of the GNU General Public License
-** along with Centreon Engine. If not, see
-** <http://www.gnu.org/licenses/>.
-*/
-
+/**
+ * Copyright 2011-2024 Centreon
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
 #ifndef CCE_OBJECTS_HOSTDEPENDENCY_HH
 #define CCE_OBJECTS_HOSTDEPENDENCY_HH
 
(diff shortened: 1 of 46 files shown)


engine/inc/com/centreon/engine/hostdependency.hh
```diff
@@ -28,18 +27,19 @@ namespace com::centreon::engine {
 class host;
 class hostdependency;
 class timeperiod;
-}
+}  // namespace com::centreon::engine
 
-typedef std::unordered_multimap<
+typedef absl::btree_multimap<
```
jean-christophe81, line 32: why not an absl::flat_hashmap (more efficient in case of std::string key)

engine/src/configuration/applier/servicedependency.cc
```diff
@@ -29,6 +29,17 @@
 
 using namespace com::centreon::engine::configuration;
 
+namespace com::centreon::engine::configuration {
+size_t servicedependency_key(const servicedependency& sd) {
+  return absl::HashOf(sd.dependency_period(), sd.dependency_type(),
```
jean-christophe81, line 34: no assert like hostdependency?

engine/src/configuration/hostescalation.cc
```diff
@@ -26,6 +26,17 @@ using namespace com::centreon::engine::configuration;
 #define SETTER(type, method) \
   &object::setter<hostescalation, type, &hostescalation::method>::generic
 
+namespace com::centreon::engine::configuration {
+
+size_t hostescalation_key(const hostescalation& he) {
+  return absl::HashOf(*he.hosts().begin(), he.contactgroups(),
```
jean-christophe81, line 32: no assert?

engine/src/configuration/applier/serviceescalation.cc
```diff
@@ -187,9 +190,10 @@ void applier::serviceescalation::remove_object(
   } else
     service_exists = true;
 
+  size_t key = serviceescalation_key(obj);
   for (serviceescalation_mmap::iterator it{range.first}, end{range.second};
        it != end; ++it) {
-    if (it->second->get_uuid() == obj.uuid()) {
+    if (it->second->internal_key() == key) {
```
jean-christophe81, line 196: event if keys are equals, objects may be differents, a first compare of key and then a compare of objects would be better.

engine/src/configuration/applier/serviceescalation.cc
```diff
@@ -250,8 +254,9 @@ void applier::serviceescalation::resolve_object(
     throw engine_error() << "Cannot find service escalations "
                          << "concerning host '" << hostname << "' and service '"
                          << desc << "'";
+  size_t key = serviceescalation_key(obj);
   for (serviceescalation_mmap::iterator it{p.first}; it != p.second; ++it) {
-    if (it->second->get_uuid() == obj.uuid()) {
+    if (it->second->internal_key() == key) {
```
jean-christophe81, line 259: idem

engine/src/configuration/applier/hostescalation.cc
```diff
@@ -246,10 +238,11 @@ void applier::hostescalation::resolve_object(
     throw engine_error() << "Cannot find host escalations concerning host '"
                          << hostname << "'";
 
+  size_t key = hostescalation_key(obj);
   for (hostescalation_mmap::iterator it{p.first}; it != p.second; ++it) {
     /* It's a pity but for now we don't have any idea or key to verify if
      * the hostescalation is the good one. */
-    if (it->second->get_uuid() == obj.uuid()) {
+    if (it->second->internal_key() == key) {
```
jean-christophe81, line 245: event if keys are equals, objects may be differents, a first compare of key and then a compare of objects would be better.

engine/src/configuration/serviceescalation.cc
```diff
@@ -29,6 +29,15 @@ using namespace com::centreon::engine::logging;
 #define SETTER(type, method) \
   &object::setter<serviceescalation, type, &serviceescalation::method>::generic
 
+namespace com::centreon::engine::configuration {
+size_t serviceescalation_key(const serviceescalation& se) {
+  return absl::HashOf(*se.hosts().begin(), *se.service_description().begin(),
```
jean-christophe81, line 34: no assert?

jean-christophe81 (approved):

PR #1465, opened 2024-06-21
Mon 34072 globals removed from conf

bouda1:
## Description

globals.hh header no more needed in the engine configuration.

REFS: MON-34072


engine/inc/com/centreon/engine/configuration/state.hh
```diff
@@ -461,6 +467,7 @@ class state {
   void use_timezone(std::string const& value);
   bool use_true_regexp_matching() const noexcept;
   void use_true_regexp_matching(bool value);
+  sched_info_config& sched_info_config() { return _scheduling_info; }
```
jean-christophe81, line 470: sched_info_config is private

bouda1: And ??
I don't understand, it works. And why I would set it public if not needed outside?

PR #1472, opened 2024-06-26
Mon 34072 daterange conf

bouda1:
## Description

Attempt to add config objects for daterange, timerange.

REFS: MON-34072

### DIFF
--- a/engine/inc/com/centreon/engine/timeperiod.hh
+++ b/engine/inc/com/centreon/engine/timeperiod.hh
@@ -79,8 +79,4 @@ void get_next_valid_time(time_t pref_time,
                          time_t* valid_time,
                          com::centreon::engine::timeperiod* tperiod);
 
-std::ostream& operator<<(std::ostream& os,
-                         com::centreon::engine::timeperiod const& obj);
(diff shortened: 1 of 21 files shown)


engine/inc/com/centreon/engine/timeperiod.hh
```diff
@@ -79,8 +79,4 @@ void get_next_valid_time(time_t pref_time,
                          time_t* valid_time,
                          com::centreon::engine::timeperiod* tperiod);
```
jean-christophe81, line 81: why remove it, it may be usefull to debug in future

engine/src/timeperiod.cc
```diff
@@ -109,41 +109,6 @@ bool timeperiod::operator!=(timeperiod const& obj) noexcept {
   return !(*this == obj);
 }
```
jean-christophe81, line 110: why remove it, it may be usefull to debug in future

bouda1: I prefer to rewrite them when needed.

bouda1: I just moved these two operators to configuration.

PR #1673, opened 2024-09-04
Update gorgone tests execution

sdepassio:
## Description

Remove the "pull" test which causes errors on el8.

### DIFF
--- a/.github/workflows/gorgone.yml
+++ b/.github/workflows/gorgone.yml
@@ -14,7 +14,7 @@ on:
       - ready_for_review
     paths:
       - "gorgone/**"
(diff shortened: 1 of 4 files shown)


.github/workflows/gorgone.yml
```diff
@@ -14,7 +14,7 @@ on:
       - ready_for_review
     paths:
       - "gorgone/**"
-      - "!gorgone/tests/**"
+      - "gorgone/tests/**"
```
tuntoja, line 17: do not forget to revert this change, as it may trigger an unexpected release if only tests are modified during a release.

PR #1682, opened 2024-09-09
fix(broker/lua): Several fix in the lua Stream

sechkem:
Broker cache not updated correctly: Several fix in the LUA Stream

REFS: MON-147529

### DIFF
--- a/broker/lua/src/macro_cache.cc
+++ b/broker/lua/src/macro_cache.cc
@@ -17,15 +17,12 @@
  */
 
 #include "com/centreon/broker/lua/macro_cache.hh"
-#include <absl/container/flat_hash_set.h>
-#include <memory>
 #include "bbdo/bam/dimension_ba_bv_relation_event.hh"
 #include "bbdo/bam/dimension_ba_event.hh"
 #include "bbdo/bam/dimension_bv_event.hh"
 #include "bbdo/storage/index_mapping.hh"
 #include "bbdo/storage/metric_mapping.hh"
 #include "com/centreon/broker/log_v2.hh"
-#include "com/centreon/broker/neb/internal.hh"
 #include "com/centreon/exceptions/msg_fmt.hh"
 
 using namespace com::centreon::exceptions;
@@ -833,7 +830,7 @@ void macro_cache::_process_host_group(std::shared_ptr<io::data> const& data) {
     } else {
       /* Here, we add the hostgroup and the first poller that needs it */
       absl::flat_hash_set<uint32_t> pollers{hg->poller_id};
-      _host_groups[hg->id] = std::make_pair(std::move(hg), pollers);
+      _host_groups[hg->id] = std::make_pair(hg, pollers);
     }
   } else {
     /* We check that no more pollers need this host group. So if the set is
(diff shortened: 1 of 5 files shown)


broker/lua/src/macro_cache.cc
```diff
@@ -17,12 +17,15 @@
  */
 
 #include "com/centreon/broker/lua/macro_cache.hh"
+#include <absl/container/flat_hash_set.h>
+#include <memory>
```
bouda1, line 21: Why are they needed?
Are you sure you need them ? I talk about flat_hash_set and memory.
We already talk about them in the header, so they at least should be declared there...

broker/lua/src/macro_cache.cc
```diff
@@ -17,12 +17,15 @@
...
 #include "bbdo/bam/dimension_ba_bv_relation_event.hh"
 #include "bbdo/bam/dimension_ba_event.hh"
 #include "bbdo/bam/dimension_bv_event.hh"
 #include "bbdo/storage/index_mapping.hh"
 #include "bbdo/storage/metric_mapping.hh"
 #include "com/centreon/broker/log_v2.hh"
+#include "com/centreon/broker/neb/internal.hh"
```
bouda1, line 28: already included in the header.

broker/lua/src/macro_cache.cc
```diff
@@ -809,14 +819,36 @@ void macro_cache::_process_pb_adaptive_host(
...
+  if (hg->enabled) {
+    auto found = _host_groups.find(hg->id);
+    if (found != _host_groups.end()) {
+      /* here, we complete the set of pollers */
+      found->second.second.insert(hg->poller_id);
+      found->second.first->name = hg->name;
+    } else {
+      /* Here, we add the hostgroup and the first poller that needs it */
+      absl::flat_hash_set<uint32_t> pollers{hg->poller_id};
+      _host_groups[hg->id] = std::make_pair(std::move(hg), pollers);
```
bouda1, line 836: Since hg is already a reference, I think you should remove the std::move

broker/lua/src/macro_cache.cc
```diff
@@ -1054,9 +1086,31 @@ void macro_cache::_process_service_group(
...
+  if (sg->enabled) {
+    auto found = _service_groups.find(sg->id);
+    if (found != _service_groups.end()) {
+      /* here, we complete the set of pollers */
+      found->second.second.insert(sg->poller_id);
+      found->second.first->name = sg->name;
+    } else {
+      /* Here, we add the servicegroup and the first poller that needs it */
+      absl::flat_hash_set<uint32_t> pollers{sg->poller_id};
+      _service_groups[sg->id] = std::make_pair(std::move(sg), pollers);
```
bouda1, line 1098: remove std::move

broker/lua/src/macro_cache.cc
sechkem: will fix

broker/lua/src/macro_cache.cc
sechkem: will fix

broker/lua/src/macro_cache.cc
sechkem: will fix

broker/lua/src/macro_cache.cc
sechkem: will fix

jean-christophe81 (approved):

bouda1 (approved):

PR #1719, opened 2024-09-24
enh(engine/gRPC): GetHost function return all about host

sechkem:
Engine gRPC API evolution : the GetHost function now return all information about the host.

REFS: MON-148834


.github/workflows/centreon-collect.yml
```diff
@@ -21,7 +21,7 @@ on:
       legacy_engine:
         description: 'Compile Engine with legacy configuration library'
         required: true
-        default: true
+        default: false
```
bouda1, line 24: Please keep the 'true' value here.

ccc/main.cc
```diff
@@ -56,6 +59,8 @@ static void usage(bool color_enabled) {
                "    Displays the available methods.\n"
                "  -n, --nocolor\n"
                "    Outputs are displayed with the current color.\n"
+               "  -F, --full\n"
+               "    Force fields with default values to always print.\n"
```
bouda1, line 63:
```suggestion
               "    Forces display of fields containing default values.\n"
```

engine/enginerpc/engine.proto
```diff
@@ -356,6 +356,122 @@ message EngineHost {
...
+  double percent_state_change = 104;
+  int32 total_services = 105;
+  uint64 total_service_check_interval = 106;
+  uint32 modified_attributes = 107;
+  int32 circular_path_checked = 108;
+  bool contains_circular_path = 109;
+  string timezone = 110;
+  uint64 icon_id = 111;
+  string group_name = 112;
+  repeated string custom_variables = 113;
```
bouda1, line 474:
```suggestion
  map<string, string> custom_variables = 113;
```

bouda1: with a map we will get a better display

engine/enginerpc/engine_impl.cc
```diff
@@ -61,12 +61,14 @@ namespace asio = boost::asio;
 #include "com/centreon/engine/servicegroup.hh"
 #include "com/centreon/engine/statistics.hh"
 #include "com/centreon/engine/statusdata.hh"
+#include "com/centreon/engine/string.hh"
```
bouda1, line 64: Can you take a look at why this header is needed.
And if it is not needed a lot, it could be good to use the absl library to make the equivalent.

engine/enginerpc/engine.proto
```diff
@@ -356,6 +356,122 @@ message EngineHost {
...
+  string child_hosts = 9;
+  string services = 10;
+  string check_command = 11;
+  State initial_state = 12;
+  uint32 check_interval = 13;
+  double retry_interval = 14;
+  int32 max_attempts = 15;
+  string event_handler = 16;
+  string contact_groups = 17;
+  string contacts = 18;
```
bouda1, line 370: Why not using repeated string here too?

engine/enginerpc/engine.proto
```diff
@@ -356,6 +356,122 @@ message EngineHost {
     UNREACHABLE = 2;
   }
   State current_state = 6;
+  string display_name = 7;
+  string parent_hosts = 8;
```
bouda1, line 360: repeated string?

engine/enginerpc/engine.proto
```diff
@@ -356,6 +356,122 @@ message EngineHost {
...
+  string child_hosts = 9;
```
bouda1, line 361: repeated strings ?

engine/enginerpc/engine.proto
```diff
@@ -356,6 +356,122 @@ message EngineHost {
...
+  string services = 10;
```
bouda1, line 362: repeated strings ?

engine/enginerpc/engine_impl.cc
```diff
@@ -234,24 +236,226 @@ grpc::Status engine_impl::GetHost(grpc::ServerContext* context [[maybe_unused]],
...
+    std::string services_str;
+    std::string notifications_str;
+    std::string groups_name_str;
+    std::string state_history_str;
+    std::string custom_variables_str;
+    std::ostringstream oss;
+    hostgroup* hg{selectedhost->get_parent_groups().front()};
+
+    if (selectedhost->get_contactgroups().empty())
+      cg_str = "\"nullptr\"";
```
bouda1, line 262: not needed, with the new option available in ccc, we can force the display of an empty field.

bouda1, line 262: not needed, with the new option available in ccc, we can force the display of an empty field.

tests/resources/Engine.py
```diff
@@ -933,47 +933,6 @@ def ctn_engine_config_delete_value_in_hosts(idx: int, desc: str, key: str, file:
...
-    Args:
-        idx (int): Index of the Engine configuration (from 0)
-        desc (str): host name of the host to modify.
-        key (str): the parameter that will be deleted.
-        file (str): The file to modify, default value 'hosts.cfg'
-    """
-    
-
-    filename = f"{ETC_ROOT}/centreon-engine/config{idx}/{file}"
-    with open(filename, "r") as f:
```
bouda1, line 945: Have you removed this function?

jean-christophe81 (approved):

engine/enginerpc/engine_impl.cc
sechkem: need for string::ctime

engine/enginerpc/engine_impl.cc
sechkem: done

sechkem: done

engine/enginerpc/engine.proto
sechkem: will fix

engine/enginerpc/engine.proto
sechkem: done

PR #1721, opened 2024-09-26
enh(engine/gRPC): GetService function return all about service

sechkem:
Engine gRPC API evolution : the GetService function now return all information about the service.

REFS: MON-148835


engine/enginerpc/engine.proto
```diff
@@ -398,6 +398,113 @@ message EngineService {
...
+  string action_url = 52;
+  string icon_image = 53;
+  string icon_image_alt = 54;
+  bool problem_has_been_acknowledged = 55;
+  enum AckType { 
+    NONE = 0; 
+    NORMAL = 1; 
+    STICKY = 2;
+  }
+  AckType acknowledgement = 56;
```
bouda1, line 455: If think there is a redundant information here
problem_has_been_acknowledged is active only if acknowledgement is NORMAL or STICKY.
So I think we can only keep the acknowledgement information.

engine/enginerpc/engine.proto
```diff
@@ -398,6 +398,113 @@ message EngineService {
...
+  string state_history = 96;
+  uint32 state_history_index = 97;
+  bool is_flapping = 98;
+  uint64 flapping_comment_id = 99;
+  double percent_state_change = 100;
+  uint32 modified_attributes = 101;
+  string host_ptr = 102;
+  string event_handler_args = 103;
+  string check_command_args = 104;
+  repeated string custom_variables = 105;
```
bouda1, line 507: protobuf defines a map that you can use here.
something like
```
map<string,string> custom_variables = 105;
```
Then we could have a view like "key=value"

engine/enginerpc/engine_impl.cc
```diff
@@ -55,17 +53,17 @@ namespace asio = boost::asio;
 #include "com/centreon/engine/globals.hh"
 #include "com/centreon/engine/hostdependency.hh"
 #include "com/centreon/engine/hostgroup.hh"
-#include "com/centreon/engine/logging/logger.hh"
 #include "com/centreon/engine/service.hh"
 #include "com/centreon/engine/servicedependency.hh"
 #include "com/centreon/engine/servicegroup.hh"
-#include "com/centreon/engine/statistics.hh"
 #include "com/centreon/engine/statusdata.hh"
+#include "com/centreon/engine/string.hh"
```
bouda1, line 60: Do you know why this header is needed?

absl library provides almost all the needed functions to manipulate strings

jean-christophe81 (approved):

PR #1740, opened 2024-10-02
Mon 116610 read vault final

bouda1:
## Description

Here is the vault configuration for Broker to get its DB credentials.

REFS: MON-116610


common/crypto/aes256.cc
```diff
@@ -0,0 +1,185 @@
...
+#include <openssl/sha.h>
+
+#include "aes256.hh"
+#include "base64.hh"
+#include "com/centreon/exceptions/msg_fmt.hh"
+
+namespace com::centreon::common::crypto {
+
+/**
+ * @brief The aes256 constructor. Th
```
jean-christophe81, line 32: brief truncated
What are the two keys?

common/crypto/aes256.cc
```diff
@@ -0,0 +1,185 @@
...
+   */
+  result.resize(iv_length + 64 + crypted_size);
+
+  std::string_view iv(result.data(), iv_length);
+  std::string_view hmac(result.data() + iv_length, 64);
+  std::string_view crypted_vector(result.data() + iv_length + 64, crypted_size);
+
+  RAND_bytes((unsigned char*)iv.data(), iv_length);
+
+  EVP_CIPHER_CTX* ctx = EVP_CIPHER_CTX_new();
```
jean-christophe81, line 68: Is it expensive to create a ctx. If yes, why not store ctx in an attribute of this class?

broker/core/sql/src/database_config.cc
```diff
@@ -93,97 +101,144 @@ database_config::database_config(const std::string& type,
...
+  if (found != cfg.params.end())
+    _password = found->second;
+
+  try {
+    common::vault::vault_access vault(env_file, vault_file, verify_peer,
+                                      _config_logger);
+    _password = vault.decrypt(_password);
+    _config_logger->info("Database password get from Vault configuration");
+  } catch (const std::exception& e) {
+    _config_logger->info("No usable Vault configuration: {}", e.what());
```
jean-christophe81, line 212: Can we avoid to display these info in case of no vault is configured (most cases)

common/http/src/https_connection.cc
```diff
@@ -142,6 +142,13 @@ https_connection::https_connection(
   }
   _stream = std::make_unique<ssl_stream>(beast::net::make_strand(*io_context),
                                          _sslcontext);
+  if (!SSL_set_tlsext_host_name(_stream->native_handle(),
```
jean-christophe81, line 145: don't do it if get_server_name() is empty

common/vault/vault_access.cc
```diff
@@ -0,0 +1,180 @@
...
+#include "com/centreon/common/http/http_config.hh"
+#include "com/centreon/common/http/https_connection.hh"
+#include "com/centreon/common/pool.hh"
+#include "com/centreon/exceptions/msg_fmt.hh"
+
+using namespace com::centreon::common::http;
+using namespace com::centreon::common::vault;
+using com::centreon::common::crypto::aes256;
+
+vault_access::vault_access(const std::string& env_file,
```
jean-christophe81, line 32: some method comments

mushroomempires (approved):
approved for owners-pipelines/dockerfile changes

common/crypto/aes256.cc
bouda1: not really. And moreover, the ctx is not initialized the same way if we encrypt or decrypt so difficult to mutualize its declaration.
I saw some people doing it but that had to declare more things in the ctx, I'm not sure it's worth.

broker/core/sql/src/database_config.cc
bouda1: I've added a filter with the password prefix.

broker/core/sql/src/database_config.cc
```diff
@@ -93,97 +100,147 @@ database_config::database_config(const std::string& type,
...
+  if (found != cfg.params.end())
+    _password = found->second;
+
+  try {
+    common::vault::vault_access vault(env_file, vault_file, verify_peer,
+                                      _config_logger);
+    _password = vault.decrypt(_password);
+    _config_logger->info("Database password get from Vault configuration");
+  } catch (const std::exception& e) {
+    const std::string_view password_prefix("secret::hashicorp_vault::");
```
jean-christophe81, line 211: constexpr

common/vault/vault_access.hh
```diff
@@ -0,0 +1,67 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ */
+#ifndef CCC_VAULT_VAULT_ACCESS_HH
+#define CCC_VAULT_VAULT_ACCESS_HH
+#include <cstdint>
```
jean-christophe81, line 20: these two hearders are mandatory, not yet included in precomp?

jean-christophe81 (approved):

PR #1812, opened 2024-10-31
Mon 151686 segfault relations

bouda1:
## Description

There was an issue with parents/children hosts relations.
When the configuration was sent to broker, we had some dangling pointers that could lead to a segfault.

REFS: MON-151686


bouda1:
@coderabbitai review

sechkem:
The Pr is obsolete, no longer needed

PR #1820, opened 2024-11-01
cleanup(code): Compilation warnings unused parameters, fallthrough and forbidden uses

sechkem:
During the compilation process, several warnings have been identified that need to be removed for more maintainable code.

REFS: MON-151231

### DIFF
--- a/clib/test/main.cc
+++ b/clib/test/main.cc
@@ -18,7 +18,7 @@
  */
 #include <gtest/gtest.h>
 
-int main(int argc, char* argv[], char** env [[maybe_unused]]) {
+int main(int argc, char* argv[]) {
   // GTest initialization.
   testing::InitGoogleTest(&argc, argv);
 
--- a/clib/test/main.cc
+++ b/clib/test/main.cc
@@ -18,7 +18,7 @@
  */
 #include <gtest/gtest.h>
 
-int main(int argc, char* argv[], char** env) {
+int main(int argc, char* argv[], char** env [[maybe_unused]]) {
   // GTest initialization.
   testing::InitGoogleTest(&argc, argv);
 
--- a/engine/src/commands/commands.cc
+++ b/engine/src/commands/commands.cc
@@ -744,10 +744,9 @@ int cmd_process_host_check_result(int cmd, time_t check_time, char* args) {
     return ERROR;
 
   ++split_it;
-  std::string output;
-  if (split_it == split.end()) {
-    output = "";
-  } else {
+
+  std::string output = "";
+  if (split_it != split.end()) {
     output = split_it->data();
     // replace \\n with \n
     string::unescape(output);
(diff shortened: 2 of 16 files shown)


clib/test/main.cc
```diff
@@ -18,7 +18,7 @@
  */
 #include <gtest/gtest.h>
 
-int main(int argc, char* argv[], char** env) {
+int main(int argc, char* argv[], char** env [[maybe_unused]]) {
```
bouda1, line 21: Isn't it possible to simply remove env argument?

engine/src/commands/commands.cc
```diff
@@ -725,28 +725,33 @@ int cmd_process_host_check_result(int cmd, time_t check_time, char* args) {
...
-    ++output;
-  } else
-    output = "";
-  int return_code(strtol(delimiter, nullptr, 0));
+  std::string host_name = std::string(*split_it);
 
-  // replace \\n with \n
-  string::unescape(output);
+  int return_code;
+  ++split_it;
```
bouda1, line 738: split_it can be split.end(), isn't it?
So you have to check this

engine/src/commands/commands.cc
```diff
@@ -725,28 +725,33 @@ int cmd_process_host_check_result(int cmd, time_t check_time, char* args) {
...
+    output = split_it->data();
+    // replace \\n with \n
+    string::unescape(output);
+  }
 
   // Submit the check result.
-  return (
-      process_passive_host_check(check_time, host_name, return_code, output));
+  return (process_passive_host_check(check_time, host_name.c_str(), return_code,
+                                     output.c_str()));
```
bouda1, line 754:
```suggestion
  return process_passive_host_check(check_time, host_name.c_str(), return_code,
                                     output.c_str());
```

clib/test/main.cc
sechkem: I will remove

engine/src/commands/commands.cc
sechkem: yes,i will fix.

bouda1 (approved):

engine/src/commands/commands.cc
```diff
@@ -725,28 +725,37 @@ int cmd_process_host_check_result(int cmd, time_t check_time, char* args) {
...
+
+  if (split_it == split.end())
+    return ERROR;
+
+  if (!absl::SimpleAtoi(*split_it, &return_code))
+    return ERROR;
+
+  ++split_it;
+  std::string output;
+  if (split_it == split.end()) {
```
jean-christophe81, line 748: replace by ```
if (split_it != split.end() {
   output = split_it->data();
    // replace \\n with \n
    string::unescape(output);
}
```

jean-christophe81 (approved):

PR #1848, opened 2024-11-07
feat(engine/grpc) : GetTag return information about a tag

sechkem:
GRPC API GetTag return information about a tag by id and type,

MON-151609

### DIFF
--- a/engine/enginerpc/engine_impl.cc
+++ b/engine/enginerpc/engine_impl.cc
@@ -942,6 +942,47 @@ grpc::Status engine_impl::GetContactGroup(grpc::ServerContext* context
 }
 
 /**
+ * @brief Return Tag informations.
+ *
+ * @param context gRPC context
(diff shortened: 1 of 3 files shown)


engine/enginerpc/engine_impl.cc
```diff
@@ -941,6 +941,47 @@ grpc::Status engine_impl::GetContactGroup(grpc::ServerContext* context
...
 }
 
+/**
+ * @brief Return Tag informations.
+ *
+ * @param context gRPC context
+ * @param request Tag's identifier (by Tag id and type)
+ * @param response The filled fields
+ *
+ *@return Status::OK
```
bouda1, line 951: I see in the code that we can send others statuses

jean-christophe81 (approved):

engine/enginerpc/engine_impl.cc
sechkem: done

PR #1858, opened 2024-11-12
fix(scope): metric names not escaped graphite 24.04

bouda1:
## Description

Metric names should be escaped now.

REFS: MON-152669

### DIFF
--- a/broker/graphite/src/query.cc
+++ b/broker/graphite/src/query.cc
@@ -299,7 +299,7 @@ void query::_get_metric_id(io::data const& d, std::ostream& is) {
  *  @param[out] is  The stream
  */
 void query::_get_metric_name(io::data const& d, std::ostream& is) {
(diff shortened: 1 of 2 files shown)


.github/workflows/package-collect.yml
```diff
@@ -221,7 +221,7 @@ jobs:
         run: rm -rf *-debuginfo*.${{ matrix.package_extension }}
 
       # set condition to true if artifacts are needed
-      - if: ${{ false }}
+      - if: ${{ true }}
```
kduret, line 224:
```suggestion
      - if: ${{ false }}
```

tuntoja (approved):
I updated the pr title so that commit message appear cleaner

jean-christophe81 (approved):

sechkem (approved):

PR #1893, opened 2024-11-26
20164 centreon-collect-24.04.7 Centreon next 24.04 (Jira release #20211#)

paloth:
REF #20164 centreon-collect-24.04.7


tuntoja:
wrong pr, closing

PR #2049, opened 2025-01-23
fix(engine) : update the icon_id for service/host when modified

sechkem:
Bad service/host update in storage.resources when modify an icon_id on service/host.

REFS: MON-157503

### DIFF
--- a/engine/src/configuration/applier/host.cc
+++ b/engine/src/configuration/applier/host.cc
@@ -298,7 +298,7 @@ void applier::host::modify_object(configuration::host const& obj) {
   // Find host object.
   host_id_map::iterator it_obj(engine::host::hosts_by_id.find(obj.key()));
   if (it_obj == engine::host::hosts_by_id.end())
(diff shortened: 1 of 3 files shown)


engine/src/configuration/applier/host.cc
```diff
@@ -298,8 +298,8 @@ void applier::host::modify_object(configuration::host const& obj) {
   // Find host object.
   host_id_map::iterator it_obj(engine::host::hosts_by_id.find(obj.key()));
   if (it_obj == engine::host::hosts_by_id.end())
-    throw engine_error() << "Could not modify non-existing "
-                         << "host object '" << obj.host_name() << "'";
+    throw engine_error() << "Could not modify non-existing " << "host object '"
+                         << obj.host_name() << "'";
```
bouda1, line 302:
```suggestion
    throw engine_error() << "Could not modify non-existing host object '"
                         << obj.host_name() << "'";
```

bouda1 (approved):

PR #2073, opened 2025-01-28
enh(ci): build docker image dedicated to packaging

kduret:
## Description

* build docker image dedicated to packaging:
  * without vcpkg : ~150MB
  * with vcpkg: reduce size from 900MB to ~500MB
* build docker image dedicated to testing
* clean vcpkg downloads after install

**Fixes** MON-158539


.github/docker/Dockerfile.packaging-centreon-collect-alma8
```diff
@@ -0,0 +1,97 @@
...
+    perl-interpreter \
+    perl-JSON \
+    perl-Thread-Queue \
+    procps-ng \
+    nfpm-2.41.1 \
+    sudo \
+    wget \
+    zstd
+
+dnf update -y libarchive
```
tuntoja, line 38: does it require an update command specifically instead of install (as I think install attempts to update a package if it already exists)

tuntoja: let s leave it as it is for now, to avoid any unforeseen consequences

tuntoja (approved):

mushroomempires (approved):

PR #2196, opened 2025-03-07
Mon 161652 dev 23 10 x broker logs somes rrd error messages seems not relevant

sechkem:
somes rrd error messages , caused by deprecated option
we removed all the data related to :
auto_rescheduling_check,auto_rescheduling_window,auto_rescheduling_interval

REFS: MON-161652


sechkem:
In version 23.10, logging is off by default, so the issue doesn't happen.

PR #2206, opened 2025-03-13
feat(ci): allow PR merge only on qa validation

kduret:
## Description

feat(ci): allow PR merge only on qa validation
This is managed by the label `qa-approved`

**Fixes** MON-158226


kduret:
Currently, `centreon-collect` does not need explicitely qa validation because everything seems automatically tested

PR #2309, opened 2025-04-22
Mon-160976 gorgone fails to install plugins on recent rh like os not containing yum

Evan-Adam:
## Description

use dnf if available when installing plugin from gorgone.

**Fixes** # MON-160976

## Type of change

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [ ] 24.04.x
- [X] master

run the script on supported OS with the correct parameter (--type=rpm or --type=deb then a centreon plugin package name )
and with a plugin package that don't exist, the package should be installed when it exist.

I remark that potential error are not transmitted to the caller of gorgone, which just check that the script was found and executed.

https://github.com/centreon/centreon-collect/blob/4281524b3d98cda884ec259c7583f108c90f0fd7/gorgone/gorgone/modules/core/action/class.pm#L244

## Checklist

- [X] I have followed the **coding style guidelines** provided by Centreon
- [X] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [X] I have commented my code, especially **hard-to-understand areas** of the PR.
- [X] I have made corresponding changes to the **documentation**.
- [X] I have **rebased** my development branch on the base branch (master, maintenance).


mushroomempires (approved):

gorgone/contrib/gorgone_install_plugins.pl
```diff
@@ -44,6 +44,10 @@
 my $command;
 if ($type eq 'rpm') {
     $command = 'yum -y install';
+    `dnf --help`;
```
bouda1, line 47: I think you can send the output to /dev/null to avoid the stdout pollution.

Or better, perl proposes things like bash, you should be able to write
```
if (-x '/usr/bin/dnf') { ... }
```
or an equivalent...

sdepassio (approved):

gorgone/contrib/gorgone_install_plugins.pl
bouda1: I have checked, and things prove I'm right:
```
my $command;
if (-x '/usr/bin/dnf') {
  $command = 'dnf -y install';
} else {
  $command = 'yum -y install';
}
```
works perfectly...

PR #2432, opened 2025-06-06
fix(gorgone): Semicolon is accepted for hostname and service name but it causes problems in DB

scresto31:
## Description

[ETL] Semicolon is accepted for hostname and service name but it causes problems in DB

**Fixes** # MON-157861

## Type of change


gorgone/tests/unit/standard/mbietl.t
```diff
@@ -0,0 +1,148 @@
...
+use lib "$FindBin::Bin/../../../";
+use gorgone::standard::library;
+use gorgone::class::db;
+use gorgone::modules::centreon::mbi::etlworkers::class;
+use gorgone::modules::centreon::mbi::etlworkers::hooks;
+use gorgone::modules::centreon::mbi::etlworkers::dimensions::main;
+use gorgone::class::module;
+use tests::unit::lib::misc;
+
+# This test uses the MySQL database created for Gorgone robot tests.
```
Evan-Adam, line 39: optionnal :
As we will probably have multiples tests using this db can you make a small func in a lib that return an hash ? 
so next time when we need either that hash or a db connexion we don't have to change this

gorgone/tests/unit/standard/mbietl.t
```diff
@@ -0,0 +1,148 @@
...
+    my %options = (
+        logger => $logger->class,
+            module_id => 'testmbietl',
+            module_id => gorgone::modules::centreon::mbi::etlworkers::hooks::NAME,
+            config_core => { gorgonecore => { internal_com_crypt => 0 } },
+            noconfig => 1,
+            data => {
+                content => {
+                    dbbi => {
+                        centstorage => { dsn => "mysql:host=$db_bi;dbname=$db_centreon_storage", user => "centreon", password=>"password" },
```
Evan-Adam, line 59: optional : 
Can you add user/password as param ?
maybe use only one host/credential for the DB, but as the password is weak it can be usefull to set our own on our env

.github/workflows/gorgone.yml
```diff
@@ -105,10 +105,63 @@ jobs:
...
+        env:
+          MYSQL_USER: centreon
+          MYSQL_PASSWORD: password
+          MYSQL_ROOT_PASSWORD: password
+
     steps:
+      - name: Install git and database client
+        shell: bash
+        run: |
+          if [[ "${{ matrix.package_extension }}" == "deb" ]]; then
```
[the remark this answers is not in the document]
scresto31: ok, moved on https://github.com/centreon/centreon-plugins/pull/5617

.github/workflows/gorgone.yml
```diff
@@ -105,10 +105,63 @@ jobs:
       credentials:
         username: ${{ secrets.HARBOR_CENTREON_PULL_USERNAME }}
         password: ${{ secrets.HARBOR_CENTREON_PULL_TOKEN }}
+
+    services:
+      mariadb:
+        image: mariadb:latest
```
kduret, line 111: you should use a tag of a version which is supported by Centreon

scresto31: Yes, in fact, I'll remove these changes and apply them directly to the unit-test Docker images https://github.com/centreon/centreon-plugins/pull/5617

gorgone/tests/unit/standard/mbietl.t
```diff
@@ -0,0 +1,148 @@
...
+# We could have set centreon-common as a dependancy for the test, but it's not that package we are testing right now, so let mock it.
+BEGIN {
+    package centreon::common::centreonvault;
+    sub get_secret {};
+    sub new {};
+    $INC{ (__PACKAGE__ =~ s{::}{/}rg) . ".pm" } = 1;
+}
+
+# same here, gorgone use a logger, but we don't want to test it right now, so we mock it.
+BEGIN {
```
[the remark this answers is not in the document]
scresto31: I added `lib/mockCentreonvault.pm` for centronvault

.github/workflows/gorgone.yml
```diff
@@ -83,7 +83,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: [unit-tests-alma8, unit-tests-alma9, unit-tests-bullseye-arm64, unit-tests-bookworm]
+        image: [unit-tests-alma8, unit-tests-alma9, unit-tests-bookworm]
```
Evan-Adam, line 86: Maybe we don't need to test arm ?

kduret (approved):

PR #2658, opened 2025-08-07
Sync hf 178754

tuntoja:
#MON-178754


tuntoja:
will be processed here => https://github.com/centreon/centreon-collect/pull/2669

PR #2680, opened 2025-08-19
Mon 180697 refactor(engine/agent): update payload schema to allow null expiration and adjust token creation logic

sechkem:
update payload schema to allow null expiration and adjust token creation logic
Refs: MON-180697

### DIFF
--- a/common/tests/jwt_test.cc
+++ b/common/tests/jwt_test.cc
@@ -176,3 +176,29 @@ TEST(TestJWT, ConstructorInvalidJsonPayload) {
     FAIL() << "Expected msg_fmt but got some other exception type.";
   }
 }
+
+// Test: never_expired
+// Purpose: Ensure that a JWT token with a null expiration time is handled
+// correctly, allowing the token to be considered valid indefinitely.
+// Note: This test checks that the expiration time is set to max when "exp" is
+// null in the payload.
+TEST(TestJWT, Never_expired) {
+  try {
+    const std::string metadata =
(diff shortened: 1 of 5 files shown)


common/tests/jwt_test.cc
```diff
@@ -176,3 +176,29 @@ TEST(TestJWT, ConstructorInvalidJsonPayload) {
...
   }
 }
+
+// Test: never_expired
+// Purpose: Ensure that a JWT token with a null expiration time is handled
+// correctly, allowing the token to be considered valid indefinitely.
+// Note: This test checks that the expiration time is set to max when "exp" is
+// null in the payload.
+TEST(TestJWT, Never_expired) {
+  try {
```
bouda1: try/catch is not needed here as google-test will catch the error for you.

common/tests/jwt_test.cc
```diff
@@ -176,3 +176,29 @@ TEST(TestJWT, ConstructorInvalidJsonPayload) {
...
+    ASSERT_EQ(jwt.get_payload(),
+              "{\"name\":\"token1\",\"iat\":1755534269,\"exp\":null}");
+    ASSERT_EQ(jwt.get_exp(), system_clock::time_point::max());
+  } catch (const msg_fmt& ex) {
+    EXPECT_TRUE(
+        std::string(ex.what()).find("forbidden values in jwt payload: document "
+                                    "doesn't respect this schema:") == 0);
+  } catch (...) {
+    FAIL() << "Expected msg_fmt but got some other exception type.";
+  }
```
bouda1: not needed too.

bouda1 (approved):

PR #2703, opened 2025-09-01
enh(grpc): support additional encryption values in configuration

sechkem:
support additional encryption values in configuration true, false , no , insecure,full
 
 REFS: MON-182272

### DIFF
--- a/engine/modules/opentelemetry/src/grpc_config.cc
+++ b/engine/modules/opentelemetry/src/grpc_config.cc
@@ -39,9 +39,11 @@ static constexpr std::string_view _grpc_config_schema(R"(
             "maximum": 65535
         },
         "encryption": {
-          "description": "encryption mode: full, insecure, or no",
-          "type": "string",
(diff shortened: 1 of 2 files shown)


engine/modules/opentelemetry/src/grpc_config.cc
```diff
@@ -126,13 +128,25 @@ grpc_config::grpc_config(const rapidjson::Value& json_config_v) {
...
-    security_mode = TLS_INSECURE;
-  } else {
-    security_mode = NONE;
+  if (json_config.has_member("encryption")) {
+    const auto& encryption_value = json_config_v["encryption"];
+    if (encryption_value.IsString()) {
+      const std::string& encryption = json_config.get_string("encryption");
+      if (encryption == "full") {
+        security_mode = TLS_SECURE;
+      } else if (encryption == "insecure") {
```
jean-christophe81: true value?

jean-christophe81 (approved):

PR #2756, opened 2025-09-17
ci(secu): backport new pipeline and gitleaks changes

sc979:
ci(secu): backport new pipeline and gitleaks changes


sc979:
Forgot that my trigger test will ping everyone... converted as draft

PR #2841, opened 2025-10-09
ci(secu): manage project tag

sc979:
ci(secu): manage project tag


sc979:
Hi @pkippes,
Same issue with the check cherry-pick failing

PR #2873, opened 2025-10-20
fix(agent): outputis truncated for cma checks

sechkem:
Output is truncated for some CMA checks

REFS: MON-185661


agent/src/check_exec.cc
```diff
@@ -158,7 +158,7 @@ void check_exec::on_completion(unsigned running_index,
   std::list<com::centreon::common::perfdata> perfs;
 
   // split multi line output
-  outputs = absl::StrSplit(std_out, absl::ByAnyChar("\r\n"), absl::SkipEmpty());
+  outputs.push_front(std_out);
```
jean-christophe81: comment to remove

agent/src/check_exec.cc
```diff
@@ -158,7 +158,7 @@ void check_exec::on_completion(unsigned running_index,
...
   if (!outputs.empty()) {
     const std::string& first_line = *outputs.begin();
     size_t pipe_pos = first_line.find('|');
```
jean-christophe81: outputs.begin()->substr(pipe_pos + 1) will returns perfdata and following lines

May be use 
```cpp
/**
 * @brief Parse buffer and fill the three strings given as references:
 *    * short_output
 *    * long_output
 *    * perf_data
 *
 * @param[in] buffer
 * @param[out] short_output
 * @param[out] long_output
 * @param[out] perf_data
 * @param[in] escape_newlines_please To escape new lines in the returned strings
 * @param[in] newlines_are_escaped To consider input newlines as escaped.
 *
 */
void parse_check_output(std::string const& buffer,
                        std::string& short_buffer,
                        std::string& long_buffer,
                        std::string& pd_buffer,
                        bool escape_newlines_please,
                        bool newlines_are_escaped)
```

jean-christophe81 (changes requested):

agent/src/check_exec.cc
```diff
@@ -157,18 +157,15 @@ void check_exec::on_completion(unsigned running_index,
...
-      std::string perfdatas = outputs.begin()->substr(pipe_pos + 1);
-      boost::trim(perfdatas);
-      perfs = com::centreon::common::perfdata::parse_perfdata(
-          0, 0, perfdatas.c_str(), _logger);
-    }
-  }
+  std::string short_output;
+  std::string long_output;
+  // parse the output check
+  parse_check_output(std_out, short_output, long_output, perfs, false, true);
```
jean-christophe81: It can't work
parse_check_output is part of engine
perfs is not a string

jean-christophe81 (changes requested):

jean-christophe81 (approved):

PR #3048, opened 2026-01-06
fix(MBI): take conditions into account when options are used for rebuild of mod_bi_metriccentilemonthlyvalue

gfontaineCentreon:
[JIRA](https://centreon.atlassian.net/browse/MON-172876)

## Type of change

## Target serie

- [ ] 23.10.x
- [ ] 24.04.x
- [x] 24.10.x
- [x] 25.10.x
- [x] develop

Run script on DB having duplicated data, or missing partitions. Data should be removed and reinsert, partitions should be created.


vuntz (approved):

gorgone/gorgone/modules/centreon/mbi/etl/perfdata/main.pm
```diff
@@ -122,45 +122,64 @@ sub purgeTables {
         $startAndEndSameMonth = 1;
     }
 
-    if ($etl->{run}->{options}->{nopurge} == 1) {
+    my $monthOnly = defined($etl->{run}->{options}->{month_only}) && $etl->{run}->{options}->{month_only} == 1;
```
omercier: There are no actual booleans in Perl, `undef` and `0` are assumed as false whereas any other value is true.
So unless `$etl->{run}->{options}->{month_only}` can take values > 1 and your want to exlude these cases, the condition `if (defined($etl->{run}->{options}->{month_only}) && $etl->{run}->{options}->{month_only} == 1)` is equivalent to `if $etl->{run}->{options}->{month_only}`.

So you could either:
- skip declaring these variables and simply put the bare original options `$etl->{run}->{options}->{any_option})` in the `if` statements below.
- keep these variables to lighten the rest of the script but simply copying the values this way `my $monthOnly = $etl->{run}->{options}->{month_only};`

vuntz (approved):

omercier (approved):

PR #3095, opened 2026-01-27
fix(notification): prevent notifications for UNKNOWN services when host goes DOWN

sechkem:
fix(engine): prevent notifications for UNKNOWN services when host goes DOWN

REFS: MON-191638

### DIFF
--- a/engine/src/service.cc
+++ b/engine/src/service.cc
@@ -1978,7 +1978,8 @@ int service::handle_async_check_result(
 
       /* (re)send notifications out about this service problem if the host is up
        * (and was at last check also) and the dependencies were okay... */
-      notify(reason_normal, "", "", notification_option_none);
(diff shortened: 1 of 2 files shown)


engine/src/service.cc
```diff
@@ -1978,7 +1978,8 @@ int service::handle_async_check_result(
 
       /* (re)send notifications out about this service problem if the host is up
        * (and was at last check also) and the dependencies were okay... */
-      notify(reason_normal, "", "", notification_option_none);
+      if (hst->get_current_state() == host::state_up)
```
jean-christophe81, line 1981: engine config flag (host down => no notify services) is not tested

jean-christophe81 (changes requested):

jean-christophe81 (approved):

PR #3130, opened 2026-02-03
enh(ci): use the standard component workflow naming pattern for colle…

tuntoja:
…ct and common

## Description

- enhance release trigger builds

Fixes #MON-191693


.github/workflows/common.yml
kduret, line 1: rename file

kduret (approved):

PR #3367, opened 2026-05-05
Mon 197731 when you move a host or a service it might be removed from cache

sechkem:
fix(broker): cache verify the instance_id before deleting

REFS: MON-197731

## Summary by Aikido
|  Security Issues: 0 |  Quality Issues: 0 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Included instance_id in protobuf acknowledgements sent by broker engine
* Added robot test that reproduced and guarded against stale deletions

**🐛 Bugfixes**
* Verified instance_id before deleting cache entries to prevent stale removals

<sup>[More info](https://app.aikido.dev/featurebranch/scan/112750058?groupId=59752)</sup>

### DIFF
--- a/broker/core/cache/src/global_cache_data.cc
+++ b/broker/core/cache/src/global_cache_data.cc
@@ -472,8 +472,7 @@ void global_cache_data::_process_pb_host(
             _logger,
             "cache: ignoring stale deletion of host {} from poller {} "
             "(currently owned by poller {})",
-            in.host_id(), in.instance_id(),
-            exist->second.first->instance_id());
+            in.host_id(), in.instance_id(), exist->second.first->instance_id());
         return;
       }
       if (exist->second.first) {
--- a/broker/core/cache/src/global_cache_data.cc
+++ b/broker/core/cache/src/global_cache_data.cc
@@ -465,6 +465,17 @@ void global_cache_data::_process_pb_host(
         _set_dirty_and_increment_modif();
       }
     } else {
+      if (in.instance_id() != 0 && exist->second.first &&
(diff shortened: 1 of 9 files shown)


broker/core/cache/src/global_cache_data.cc
```diff
@@ -465,6 +465,16 @@ void global_cache_data::_process_pb_host(
         _set_dirty_and_increment_modif();
       }
     } else {
+      if (in.instance_id() != 0 && exist->second.first &&
```
jean-christophe81, line 468: in.instance_id() can happen?

broker/core/cache/src/global_cache_data.cc
```diff
@@ -800,6 +810,17 @@ void global_cache_data::_process_pb_service(
...
```
jean-christophe81, line 813: same

jean-christophe81 (changes requested):

jean-christophe81 (approved):

broker/core/cache/src/global_cache_data.cc
sechkem: If the instance_id field is missing from the message, protobuf sets it to its default value (0). For a real poller, instance_id must be non-zero.

broker/core/cache/src/global_cache_data.cc
sechkem: If the instance_id field is missing from the message, protobuf sets it to its default value (0). For a real poller, instance_id must be non-zero.

PR #3368, opened 2026-05-05
fix(packaging): add missing LUA-version

sechkem:
fix(packaging): add missing LUA-version

REFS: MON-198587


kduret (commented):
centreon-collect is not packaged anymore fro bookworm, jammy and noble

kduret (approved):
approved to avoid blocking release. But we'll need to check why centreon-broker-core is needed for cma

tuntoja (approved):

PR #3457, opened 2026-06-12
MON-200794 fix(engine): timeperiod inheritance - preserve child exceptions and days during template merge

jean-christophe81:
REFS:MON-200794
## Summary

- Fix template merge for `ExceptionArray`: template exceptions are now merged into child (instead of overwriting child's own exceptions)
- Fix template merge for `DaysArray`: only inherit weekly time ranges from template for days the child does not define (instead of overwriting all child ranges)
- Add `TIMEPERIOD_MERGE` robot test validating that a child timeperiod's own calendar exception survives the merge with a template

## Root cause

In `parser::_merge()`, `ExceptionArray` and `DaysArray` fields were falling through to the generic `CopyFrom` branch, which replaced the child's entire field with the template's. This caused:
- A child timeperiod's own exceptions to be silently dropped if the template had any exceptions
- A child's custom weekday ranges to be overwritten by the template's ranges

## Test Plan

- [ ] `TIMEPERIOD_INHERITANCE` — child inherits template exception (future window) → service NOT checked
- [ ] `TIMEPERIOD_INHERITANCE_EXCEPTION` — child 24x7 ranges take priority over template future exception → service IS checked
- [ ] `TIMEPERIOD_MERGE` (new) — child's own calendar exception (active now) is preserved after merge with template that has a different exception (yesterday) → service IS checked

### DIFF
--- a/common/engine_conf/parser.cc
+++ b/common/engine_conf/parser.cc
@@ -652,13 +652,11 @@ void parser::_resolve_template(std::unique_ptr<message_helper>& msg_helper,
   if (days->weekday().empty() && !template_days->weekday().empty()) { \
     days->mutable_##weekday()->Add(template_days->weekday().begin(),  \
                                    template_days->weekday().end());   \
-    msg_helper->set_changed(f->index());                              \
   }
 
 template <class repeated_Daterange_type>
-static bool _merge_exceptions(repeated_Daterange_type* except,
+static void _merge_exceptions(repeated_Daterange_type* except,
                               const repeated_Daterange_type& template_except) {
-  bool at_least_one_added = false;
   for (const Daterange& to_insert : template_except) {
     bool found = false;
     for (const Daterange& to_check : *except) {
@@ -670,10 +668,8 @@ static bool _merge_exceptions(repeated_Daterange_type* except,
     }
     if (!found) {
       except->Add()->CopyFrom(to_insert);
-      at_least_one_added = true;
     }
   }
-  return at_least_one_added;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


jean-christophe81:
### Code review

Found 3 issues:

1. **Debug leftover: `ShortDebugString()` instead of `timeperiod_name()`** — dumps the full protobuf text (all exceptions, all day ranges) for every resolved timeperiod. All other log lines in this file use `obj.timeperiod_name()`.

https://github.com/centreon/centreon-collect/blob/7f72dfd70247cc4789c2769f82dbb7ec9dd1f755/engine/src/configuration/applier/timeperiod.cc#L149-L153

2. **Debug leftover: all 5 log levels set to `trace` in `TIMEPERIOD_INHERITANCE_EXCEPTION`** — these were not cleaned up before the PR. No other test in the suite does this, and the companion `TIMEPERIOD_MERGE` test doesn't set any of them. This will produce very verbose output and slow CI runs.

https://github.com/centreon/centreon-collect/blob/7f72dfd70247cc4789c2769f82dbb7ec9dd1f755/tests/engine/timeperiod_inheritance.robot#L54-L60

3. **Inverted failure message in `TIMEPERIOD_INHERITANCE_EXCEPTION`** — `Should Be True` (asserts service IS checked) but the failure message reads `"service_1 must remained unknown and not checked"`. Copy-pasted from the `Should Not Be True` in the test above. If this assertion ever fails, the error message will actively mislead.

https://github.com/centreon/centreon-collect/blob/7f72dfd70247cc4789c2769f82dbb7ec9dd1f755/tests/engine/timeperiod_inheritance.robot#L84-L87

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sechkem (approved):

PR #3486, opened 2026-06-26
feat(docker): support custom plugins and dependencies in centreon-engine container

Tpo76:
## MON-201883

Adds two mechanisms to let users extend the centreon-engine container with their own plugins and dependencies, without modifying the base image.

### Custom scripts — volume mount point
A dedicated directory `/usr/lib/nagios/plugins/custom/` is created and owned by `centreon-engine` in the Dockerfile. Users mount their scripts there via docker-compose:
```yaml
volumes:
  - ./my-plugins:/usr/lib/nagios/plugins/custom:ro
```
Scripts must be executable (`chmod +x`).

### Runtime APT dependencies — `custom-deps.json`
New entrypoint script `41-custom-deps.sh` (runs after `40-engine.sh`) reads `/etc/centreon-engine/custom-deps.json` at startup and installs the listed APT packages:
```json
{
  "apt": ["snmp", "python3-requests", "libnet-snmp-perl"]
}
```
Package names are validated against Debian naming rules in `check_custom_deps.py` before being passed to `apt-get`.

Hot-reload is handled by `46-custom-deps-watcher_background.sh`: same inotify pattern as the existing `45-plugin-watcher_background.sh`, re-installs whenever `custom-deps.json` changes.

Mount in docker-compose:
```yaml
volumes:
  - ./custom-deps.json:/etc/centreon-engine/custom-deps.json:ro
```

No sudoers changes required — `apt`/`apt-get` are already allowed.

### Custom image build
For pip3/cpan/compiled dependencies, users extend the published image:
```dockerfile
FROM centreon/centreon-engine:24.10
USER root
RUN apt-get update && apt-get install -y --no-install-recommends \
    python3-boto3 libnet-snmp-perl && \
    rm -rf /var/lib/apt/lists/*
USER centreon-engine
```

## Test plan
- [x] Build with existing `.deb` mount → unchanged behaviour
- [x] Mount a script directory → verify centreon-engine can call `/usr/lib/nagios/plugins/custom/my-script.sh`
- [x] Mount `custom-deps.json` with `"apt": ["snmp"]` → `dpkg -l snmp` shows installed after startup
- [x] Modify `custom-deps.json` while container is running → package installs without restart
- [ ] Invalid package name in `custom-deps.json` → skipped with warning, valid packages still installed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/docker/centreon-engine/trixie/entrypoint/container.d/41-custom-deps.sh
+++ b/.github/docker/centreon-engine/trixie/entrypoint/container.d/41-custom-deps.sh
@@ -3,20 +3,17 @@
 set -e
 
 DEPS_JSON="/etc/centreon-engine/custom-deps.json"
-APT_LOCK="/var/lock/centreon-engine-apt.lock"
 
 [ -f "$DEPS_JSON" ] || exit 0
 
-(
-    flock -w 120 9 || { echo "apt lock timeout, skipping custom deps install"; exit 0; }
-    export DEBIAN_FRONTEND=noninteractive
-    PKGS=$(python3 /var/lib/centreon-engine/check_custom_deps.py "$DEPS_JSON" 2>/dev/null)
-    if [ -n "$PKGS" ]; then
-        echo "Installing custom APT deps: $PKGS"
-        sudo apt-get update -qq > /proc/1/fd/1 2>&1 || true
-        # shellcheck disable=SC2086
-        sudo apt-get install -y -- $PKGS > /proc/1/fd/1 2>&1 || true
-    else
-        echo "custom-deps.json found but no valid APT packages to install"
-    fi
-) 9>"${APT_LOCK}" &
+. /var/lib/centreon-engine/apt_install.sh
+
+PKGS=$(python3 /var/lib/centreon-engine/check_custom_deps.py "$DEPS_JSON" 2>/dev/null)
+if [ -n "$PKGS" ]; then
+    echo "Installing custom APT deps: $PKGS"
+    # Run in background so centengine starts without waiting for apt.
+    # shellcheck disable=SC2086
+    apt_install_pkgs $PKGS &
+else
+    echo "custom-deps.json found but no valid APT packages to install"
+fi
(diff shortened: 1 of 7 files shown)


.github/docker/centreon-engine/trixie/entrypoint/container.d/41-custom-deps.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+set -e
```
tuntoja: shouldn t we keep the "set" at parent script level ? container.sh already has this and children should not change parents behavior I think

.github/docker/centreon-engine/trixie/entrypoint/container.d/41-custom-deps.sh
```diff
@@ -0,0 +1,19 @@
...
+
+DEPS_JSON="/etc/centreon-engine/custom-deps.json"
+
+[ -f "$DEPS_JSON" ] || exit 0
```
tuntoja: wouldn t an exit 0 here cause the parent script to exit as well ? (since the numbered scripts are sourced)

.github/docker/centreon-engine/trixie/entrypoint/container.d/41-custom-deps.sh
Tpo76: Good catch, since scripts are sourced, `set -e` leaks into the parent and container.sh already sets it. Removed from 40 and 41. Fixed in cff184cdf.

.github/docker/centreon-engine/trixie/entrypoint/container.d/41-custom-deps.sh
Tpo76: Exactly `exit 0` here killed container.sh (PID 1) before centengine started, whenever custom-deps.json wasn't mounted. Switched to `return 0`. Fixed in cff184cdf.

tuntoja (approved):

PR #3526, opened 2026-07-06
fix(ci): promote docker floating tag only after e2e runtime tests pass

Tpo76:
## Summary

Fixes MON-203216: `dockerize-engine-broker` (`collect.yml`) and `dockerize` (`docker-gorgone.yml`) pushed the floating `develop` tag in the **same** `docker/build-push-action` step as the versioned candidate tag, before `docker-boot-test`/`docker-config-wiring-test` ran. A failed test left `centreon-engine-trixie:develop` (or `centreon-broker-trixie:develop`, `centreon-gorgone-trixie:develop`) pointing at a broken image until the next green nightly.

- Build/push step now pushes **only** the versioned candidate tag (`develop-<major>`).
- New `promote-docker-tag` job runs `docker buildx imagetools create` (manifest retag, no rebuild) to point the floating tag at that same digest — gated on `docker-boot-test` **and** `docker-config-wiring-test` both succeeding, and skipped entirely when there's no floating tag to promote (PRs, `dev-*.x`).
- `centreon-broker` has no e2e tests of its own; its floating-tag promotion is gated on the `centreon-engine` tests (same build, same job) — imperfect but strictly better than the current no-guard-at-all state. Adding broker-specific e2e coverage is left for a follow-up ticket.
- `docker-gorgone.yml`'s `docker-boot-test`/`docker-config-wiring-test` were also missing the explicit `needs.*.result == 'success'` guard that `collect.yml` already carries (worked around a GitHub Actions implicit-`success()` skip bug, see commit `dc7d714ba`). Added it here too since the new promote job now depends on these tests actually running.

**Verification note:** confirmed no in-repo consumer (in `centreon-collect` or `centreon`) references the floating `:develop` tag for these three images — the boot/wiring test jobs already pull the immutable per-run `image_tag`. The floating tag's consumers are external (per the ticket's stated risk), which this fix protects.

## Test plan

- [x] `actionlint` clean on both modified files (no `-ignore` overrides needed beyond the repo's existing set)
- [x] `yamllint` clean with the repo's exact CI rule set
- [x] On this PR: confirm `promote-docker-tag` appears in the Actions graph and is **skipped** (branch name != `develop`, so `additional_tag` is empty) — expected no-op
- [ ] After merge, on the next `develop` push/nightly: confirm `promote-docker-tag` runs after both test jobs go green, and that `centreon-engine-trixie:develop` / `centreon-broker-trixie:develop` / `centreon-gorgone-trixie:develop` end up pointing at the same digest as `...:develop-<major>`

### DIFF
--- a/.github/workflows/collect.yml
+++ b/.github/workflows/collect.yml
@@ -666,10 +666,8 @@ jobs:
     needs: [get-environment, dockerize-engine-broker, docker-boot-test, docker-config-wiring-test]
     if: |
       ! cancelled() &&
-      needs.get-environment.result == 'success' &&
-      needs.dockerize-engine-broker.result == 'success' &&
-      needs.docker-boot-test.result == 'success' &&
-      needs.docker-config-wiring-test.result == 'success' &&
+      ! contains(needs.*.result, 'failure') &&
+      ! contains(needs.*.result, 'cancelled') &&
       needs.dockerize-engine-broker.outputs.additional_tag != ''
     runs-on: ubuntu-24.04
     strategy:
--- a/.github/workflows/collect.yml
+++ b/.github/workflows/collect.yml
@@ -467,6 +467,7 @@ jobs:
     # matrix.image or matrix.distrib, so every leg produces the same tag.
     outputs:
       image_tag: ${{ steps.compute-tag.outputs.tag }}
+      additional_tag: ${{ steps.compute-tag.outputs.additional_tag }}
 
     steps:
       - name: Checkout sources
@@ -560,7 +561,6 @@ jobs:
           push: true
           tags: |
             ${{ vars.DOCKER_INTERNAL_REGISTRY_URL }}/${{ matrix.image }}-${{ matrix.distrib }}:${{ steps.compute-tag.outputs.tag }}
-            ${{ steps.compute-tag.outputs.additional_tag != '' && format('{0}/{1}-{2}:{3}', vars.DOCKER_INTERNAL_REGISTRY_URL, matrix.image, matrix.distrib, steps.compute-tag.outputs.additional_tag) || '' }}
           secrets: |
             "ARTIFACTORY_INTERNAL_REPO_USERNAME=${{ secrets.ARTIFACTORY_INTERNAL_REPO_USERNAME }}"
             "ARTIFACTORY_INTERNAL_REPO_PASSWORD=${{ secrets.ARTIFACTORY_INTERNAL_REPO_PASSWORD }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-gorgone.yml
```diff
@@ -182,13 +183,16 @@ jobs:
...
           secrets: |
             "ARTIFACTORY_INTERNAL_REPO_USERNAME=${{ secrets.ARTIFACTORY_INTERNAL_REPO_USERNAME }}"
             "ARTIFACTORY_INTERNAL_REPO_PASSWORD=${{ secrets.ARTIFACTORY_INTERNAL_REPO_PASSWORD }}"
 
   docker-boot-test:
     needs: [get-environment, dockerize]
+    if: |
+      ! cancelled() &&
+      needs.get-environment.result == 'success' &&
+      needs.dockerize.result == 'success'
```
kduret:
```suggestion
      ! cancelled() &&
      ! contains(needs.*.result, 'failure') &&
      ! contains(needs.*.result, 'cancelled')
```

.github/workflows/collect.yml
```diff
@@ -662,6 +662,39 @@ jobs:
...
           retention-days: 1
 
+  promote-docker-tag:
+    needs: [get-environment, dockerize-engine-broker, docker-boot-test, docker-config-wiring-test]
+    if: |
+      ! cancelled() &&
+      needs.get-environment.result == 'success' &&
+      needs.dockerize-engine-broker.result == 'success' &&
+      needs.docker-boot-test.result == 'success' &&
+      needs.docker-config-wiring-test.result == 'success' &&
```
kduret:
```suggestion
      ! cancelled() &&
      ! contains(needs.*.result, 'failure') &&
      ! contains(needs.*.result, 'cancelled') &&
```

.github/workflows/docker-gorgone.yml
Tpo76: Applied, thanks. 
Also aligned docker-config-wiring-test and promote-docker-tag in this same file for consistency (both had the identical explicit-check pattern). Safe here specifically because dockerize in docker-gorgone.yml has no if: of its own. 
It never skips, only succeeds or fails, so contains(needs.*.result, 'failure'/'cancelled') is equivalent to the explicit check.

.github/workflows/collect.yml
Tpo76: Applied, thanks. 
Also aligned the equivalent promote-docker-tag job in docker-gorgone.yml. Kept needs.dockerize-engine-broker.outputs.additional_tag != '' as-is: that's the real guard against running when dockerize-engine-broker was skipped (nightly/label gating), independent of the failure/cancelled check. Left collect.yml's docker-boot-test/docker-config-wiring-test untouched on purpose.
dockerize-engine-broker DOES self-skip there with no equivalent additional_tag-style guard, so the strict needs.dockerize-engine-broker.result == 'success' check is still required to avoid running against a skipped build.

kduret (approved):

PR #3527, opened 2026-07-06
feat(ci): promote Docker images to Pulp on stable release only

Tpo76:
## Summary

MON-204486: deliver Docker images to Pulp (our future public registry) as a **stable-only promote from Harbor**, not a build-time mirror.

**Depends on #3526 (MON-203216)** — stacked on top of it (base branch retargeted). Reuses its `image_tag`/`additional_tag` job outputs and its e2e-gated `promote-docker-tag` job structure as a model.

## Design: 3-tier delivery, Harbor internal / Pulp public

Harbor stays the sole home for candidate and testing images (unchanged from #3526). Pulp only ever receives a **final stable release**:

- `dockerize-engine-broker` (`collect.yml`) / `dockerize` (`docker-gorgone.yml`) now also build (and push to Harbor, tagged with the sanitized branch name, e.g. `release-24.10-next`) when `stability == 'testing'` — i.e. on every push to a `release-*-next`/`hotfix-*-next` branch, the same trigger `deliver-rpm`/`deliver-deb` already use. Previously these jobs only built on nightly cron or an explicit label, so no Docker image ever existed for a release branch's exact tip.
- New `promote-docker-to-pulp` job, gated on `stability == 'stable'` (fires when the per-component release tag — `centreon-collect-XX.YY.Z` / `centreon-gorgone-XX.YY.Z` — is pushed by `release-new`). It reconstructs the Harbor tag the testing build produced (`<release_type>-<major_version>-next`, where `release_type` comes from the annotated tag's message, not the tag's own ref) and copies that image straight to Pulp under the real release version (`<major>.<minor>` and floating `<major>`) via `docker buildx imagetools create` — no rebuild.
- `docker-push-to-pulp.sh` is now single-purpose (promote only, the old build-time `mirror` mode is gone): it also strips the OCI attestation/referrers manifest before publishing, since Pulp's container registry deterministically 500s when committing a manifest that includes it (confirmed via a `provenance:false` before/after test). Harbor keeps full provenance; only the Pulp copy omits it.
- The whole Pulp step is step-level `continue-on-error: true`, same as before, but now with a visible `::warning::` when it fails — a missed *stable* delivery must never be silent.
- `docker-gorgone.yml` never triggered on `master`/`XX.YY.x`/release/hotfix branches or tags at all before this PR — added those triggers (mirroring `collect.yml`'s existing `on.push`) so the promote job can fire.
- Reverted the runner (`ubuntu-24.04`) and `permissions` on jobs that no longer talk to Pulp directly (build jobs, `promote-docker-tag`'s Harbor floating-tag step); only `promote-docker-to-pulp` runs on `centreon-ubuntu-22.04` with `id-token: write`.

## Why this shape (vs. the reference #10842 / earlier iterations of this PR)

Earlier versions of this PR mirrored Pulp at build time and re-promoted the nightly floating tag — decoupled and non-blocking, but it meant Pulp tracked `develop`'s nightly churn, which doesn't demonstrate anything about a production delivery path. The target architecture (Harbor internal, Pulp public, JFrog retired) needs Pulp to only ever hold real releases. This PR re-scopes to that: candidate/testing traffic never leaves Harbor, Pulp is populated once, at the moment a version actually goes stable — same shape as the RPM/DEB `deliver` → `promote` flow.

## Known limitations (tracked for follow-up, not blocking this delivery)

- The reconstructed Harbor tag assumes the testing build for that release/hotfix branch actually ran and is still present on Harbor when the stable tag is pushed. There's no verification step beyond `imagetools inspect` failing (caught, warned, non-blocking) — confirmed working as designed (see test plan below): a missing Harbor tag fails cleanly with a visible warning, doesn't block the job.
- The automatic trigger chain itself (testing build on a `release-*-next`/`hotfix-*-next` push → branch-name Harbor tag → stable component tag push → `release_type`/`major_version` reconstruction) has not yet run end-to-end against a real release — only the `promote-docker-to-pulp` mechanics were exercised directly (see below). First live proof will be the next real (or deliberately staged) release/hotfix cut.

## Scope decision

Same three images as #3526: `centreon-engine`, `centreon-broker`, `centreon-gorgone`. `centreon-poller` (no floating tag, no tests) stays untouched.

## Test plan

- [x] `actionlint` (repo's exact CI ignore ruleset) clean on both workflows
- [x] `bash -n` syntax check on the script
- [x] Verified `MAJOR` is consistent across `.version`, `CMakeLists.txt` (`COLLECT_MAJOR`/`COLLECT_MINOR`), and `.version.centreon-gorgone` (only overrides `MINOR`) — the Harbor tag reconstruction relies on this
- [x] `promote-docker-to-pulp` validated end-to-end for all three images via a temporary `workflow_dispatch` bypass input (added, exercised, then removed once green — same throwaway pattern as #3526's `simulate_promote_tag`): resolved per-platform Harbor manifests, stripped the attestation manifest, `imagetools create`'d to Pulp, and PATCHed labels, all under a synthetic `test-promote-<run_id>` tag (never a real version). Runs [28804318369](https://github.com/centreon/centreon-collect/actions/runs/28804318369) (engine + broker) and [28803918541](https://github.com/centreon/centreon-collect/actions/runs/28803918541) (gorgone). Caught and fixed a real bug in the process: the new job was missing its `actions/checkout` step (script wasn't present on the runner, exit 127).
- [x] Confirmed the non-blocking/warning behavior on a genuinely missing Harbor tag (first attempt used a stale `develop-26.09` reference that didn't exist yet for engine/broker — failed cleanly with `::warning::`, job stayed green, no CI impact)
- [ ] End-to-end validation of the *automatic* trigger chain (testing-branch build → stable tag push → promote) against a real or deliberately staged release/hotfix cut


pkriko (dismissed):

.github/workflows/docker-gorgone.yml
```diff
@@ -10,11 +10,17 @@ on:
     branches:
       - develop
       - dev-[2-9][0-9].[0-9][0-9].x
+      - master
```
tuntoja: master can be dropped, even if it is still visible, it does not receive any more code

Tpo76: Dropped, thanks.

.github/workflows/docker-gorgone.yml
tuntoja, line 209: if L102 is applied and skips dockerized on stable stability, it means the dockerize ouputs would be empty, since a skip job is seen a success by github

tuntoja, line 262: if L102 is applied and skips dockerized on stable stability, it means the dockerize ouputs would be empty, since a skip job is seen a success by github

.github/workflows/docker-gorgone.yml
```diff
@@ -311,3 +317,76 @@ jobs:
             --tag "${{ vars.DOCKER_INTERNAL_REGISTRY_URL }}/centreon-gorgone-${{ matrix.operating_system }}:${{ needs.dockerize.outputs.additional_tag }}" \
             "${{ vars.DOCKER_INTERNAL_REGISTRY_URL }}/centreon-gorgone-${{ matrix.operating_system }}:${{ needs.dockerize.outputs.image_tag }}"
         shell: bash
+
+  promote-docker-to-pulp:
+    needs: [get-environment]
+    if: |
```
tuntoja, line 322: this may need the same conditions as the package promote skip_workflow == 'false' and github.repository == 'centreon/centreon-collect'

tuntoja (approved):

pkriko (approved):

PR #3563, opened 2026-07-16
MON-169103-hostgroups-disappear

jean-christophe81:
REFS:MON-169103
fix(broker): when a host, service, group member is deleted by a poller we insure that another poller is not the owner

### Problem

When a host, a service, or a group membership (host group / service group) is moved from one poller to another, the two pollers reload independently and send their events in an unpredictable order:

* the **old poller** sends deletion/disable events for the host, service, or group membership it no longer manages,
* the **new poller** sends creation/enable events for the same objects, now attached to it.

Since both pollers write to the same `centreon_storage` tables, if the deletion event from the old poller arrives **after** the creation event from the new poller, it deletes the row that the new poller just (re)created. The host, service, or group membership then disappears permanently from the database, even though it is still configured and monitored by the new poller.

### Fix

The fix makes every deletion/disable query poller-aware, so a poller can only delete or disable data that it currently owns (matched on `instance_id` / `poller_id`), instead of matching only on the object's functional key (`host_id`, `service_id`, `hostgroup_id`, `servicegroup_id`):

* **Host group / service group membership deletion** (`_process_host_group_member`, `_process_pb_host_group_member`, `_process_service_group_member`, `_process_pb_service_group_member` in `broker/unified_sql/src/stream_sql.cc`): the `DELETE` statements now join on the `hosts` table and filter on `instance_id`/`poller_id`, so a membership is only removed if it still belongs to the poller sending the deletion.
* **Host and service disabling** (`_process_host`, `_process_pb_host`, `_process_pb_service`): disabling a host or service (i.e. `enabled = 0`) now goes through a dedicated `UPDATE` statement keyed on `(host_id, instance_id)` / `(host_id, service_id, instance_id)` instead of the generic insert-or-update statement, so a poller can't disable a host/service that has already been re-created by another poller. New prepared statements `_host_update`, `_pb_host_update`, `_pb_service_update` were added for this, along with a shared `_prepare_pb_requests()` helper to prepare all the protobuf statements together.
* **Resources table cleanup** (`_process_pb_host_in_resources`, `_process_pb_service_in_resources`): the `resources` disabling query now also filters on `poller_id`.
* `query_preparator::prepare_update_table` was adjusted to build the `WHERE` clause from the protobuf unique-key fields (`_pb_unique`) instead of skipping them, so the new poller-aware `UPDATE` statements can be generated correctly.

### Tests

Added two Robot Framework scenarios reproducing the race condition on a 2-poller setup, moving hosts/services from poller 0 to poller 1 and reloading poller 1 before poller 0 (the ordering that used to trigger the bug):
* `EBNHG6` in `tests/broker-engine/hostgroups.robot`: moves `host_1` and `host_2` (members of hostgroup 1) to another poller and checks the hostgroup and its host memberships survive.
* `EBNSG2` in `tests/broker-engine/servicegroups.robot`: moves `host_1`/`service_1` and `host_2`/`service_2` (members of servicegroup 1) to another poller and checks the servicegroup and its service memberships survive.

**Fixes** MON-169103

## Type of change

## Target serie

- [ ] 23.10.x
- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [ ] master

Run the new Robot Framework tests:
```
robot -i MON-169103 tests/broker-engine/hostgroups.robot tests/broker-engine/servicegroups.robot
```
Both `EBNHG6` and `EBNSG2` set up two pollers, attach hosts/services to a group on poller 0, move them to poller 1, then reload poller 1 before poller 0 (the ordering that previously caused the deletion event from poller 0 to arrive after the creation event from poller 1). The group and its memberships must still be present in `hostgroups`/`hosts_hostgroups` (resp. `servicegroups`/`services_servicegroups`) after both reloads complete.

## Checklist

- [ ] I have followed the **coding style guidelines** provided by Centreon
- [ ] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [ ] I have commented my code, especially **hard-to-understand areas** of the PR.
- [ ] I have made corresponding changes to the **documentation**.
- [ ] I have **rebased** my development branch on the base branch (master, maintenance).

### DIFF
--- a/broker/unified_sql/src/stream_sql.cc
+++ b/broker/unified_sql/src/stream_sql.cc
@@ -1425,16 +1425,14 @@ void stream::_process_host_group_member(const std::shared_ptr<io::data>& d) {
         "SQL: disabling membership of host {} to host group {} on instance {}",
         hgm.host_id, hgm.group_id, hgm.poller_id);
 
-    if (!_host_group_member_delete.prepared()) {
-      query_preparator::event_unique unique;
-      unique.insert("hostgroup_id");
-      unique.insert("host_id");
-      query_preparator qp(neb::host_group_member::static_type(), unique);
-      _host_group_member_delete = qp.prepare_delete(_mysql);
-    }
-    _host_group_member_delete << hgm;
-    _mysql.run_statement(_host_group_member_delete,
-                         database::mysql_error::delete_host_group_member, conn);
+    std::string query = fmt::format(
+        "DELETE hosts_hostgroups FROM hosts_hostgroups INNER JOIN hosts ON "
+        "hosts_hostgroups.host_id=hosts.host_id "
+        "WHERE hosts.host_id={} and hostgroup_id = {} and instance_id = {}",
+        hgm.host_id, hgm.group_id, hgm.poller_id);
+
+    _mysql.run_query(query, database::mysql_error::delete_host_group_member,
+                     conn);
     _add_action(conn, actions::hostgroups);
   }
 }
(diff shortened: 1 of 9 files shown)


broker/unified_sql/src/stream_sql.cc
```diff
@@ -4168,8 +4245,8 @@ uint64_t stream::_process_pb_service_in_resources(const Service& s,
       SPDLOG_LOGGER_INFO(
           _logger_sql,
           "SQL: no need to remove service ({}, {}), it is not in "
-          "database",
-          s.host_id(), s.service_id());
+          "database or is not owned py poller {}",
```
sechkem:
```suggestion
          "database or is not owned by poller {}",
```

sechkem (dismissed):

sechkem (approved):

PR #3575, opened 2026-07-21
enh(broker): add icon info 

sechkem:
enh(broker): add icon info.
REFS: MON-197568.

### DIFF
--- a/agent/native_windows/src/check_cpu.cc
+++ b/agent/native_windows/src/check_cpu.cc
@@ -398,7 +398,8 @@ check_cpu::check_cpu(const std::shared_ptr<asio::io_context>& io_context,
           if (val) {
             _use_nt_query_system_information = *val;
           }
-        } else if (member_iter->name != "cpu-detailed") {
+        } else if (member_iter->name != "cpu-detailed" &&
+                   member_iter->name != "timeout") {
           SPDLOG_LOGGER_ERROR(logger, "command: {}, unknown parameter: {}",
                               get_command_name(), member_iter->name);
         }
--- a/agent/native_windows/src/check_memory.cc
+++ b/agent/native_windows/src/check_memory.cc
@@ -444,6 +444,9 @@ check_memory::check_memory(const std::shared_ptr<asio::io_context>& io_context,
         }
         continue;
       }
+      if (key == "timeout") {
+        continue;
+      }
 
       auto mem_to_status_search = _label_to_mem_to_status.find(key);
       if (mem_to_status_search != _label_to_mem_to_status.end()) {
--- a/agent/native_windows/src/check_service.cc
+++ b/agent/native_windows/src/check_service.cc
@@ -802,7 +802,7 @@ check_service::check_service(
       } else if (key != "filter-name" && key != "exclude-name" &&
                  key != "filter-display" && key != "exclude-display" &&
                  key != "start-auto" && key != "start-type" &&
-                 key != "delayed" && key != "type") {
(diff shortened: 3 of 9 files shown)


jean-christophe81 (dismissed):

vuntz:
_This comment was written with the help of AI, but guided and reviewed by me_

I looked at this PR because we want to use this new info in centreon-map. I think there is a problem with `icon_image`: the directory is missing.

In `broker/bam/src/configuration/reader_v2.cc` the query only joins `view_img`:

```sql
LEFT JOIN view_img AS vi ON vi.img_id=b.icon_id
```

So `icon_image` contains only `view_img.img_path`, for example `logo.png`. But `img_path` is always just the file name, the directory is in `view_img_dir_relation` and `view_img_dir`. You can see it in centreon-web when it inserts the media:
[`syncDir.php#L231-L245`](https://github.com/centreon/centreon/blob/develop/centreon/www/include/options/media/images/syncDir.php#L231-L245).

For normal services, the value contains the directory, because centreon-web resolves it when it generates the engine configuration:

- [`media.class.php#L38-L50`](https://github.com/centreon/centreon/blob/develop/centreon/www/class/config-generate/media.class.php#L38-L50): `getMediaPathFromId()` returns `dir_name . '/' . img_path`
- [`abstract/service.class.php#L140-L143`](https://github.com/centreon/centreon/blob/develop/centreon/www/class/config-generate/abstract/service.class.php#L140-L143): it writes this in `services.cfg` as `icon_image`
- `engine/src/configuration/applier/service.cc:302`: engine keeps the value as it is
- `engine/src/broker.cc:1086-1088`: broker copies the value as it is in `PB_SERVICE`

So a normal service sends `centreon/logo.png`, but a BA will send only `logo.png`. Consumers add the media folder prefix (`/img/media/`) to this value, so for a BA we get `/img/media/logo.png` and the image is not found.

I think adding the two missing joins is enough:

```sql
SELECT b.ba_id, b.name, b.state_source, b.level_w,
       b.level_c, b.last_state_change, b.current_status,
       b.in_downtime, b.inherit_kpi_downtimes, b.icon_id,
       CONCAT(vid.dir_name, '/', vi.img_path), vi.img_comment
FROM mod_bam AS b
INNER JOIN mod_bam_poller_relations AS pr ON b.ba_id=pr.ba_id
LEFT JOIN view_img AS vi ON vi.img_id=b.icon_id
LEFT JOIN view_img_dir_relation AS vidr ON vidr.img_img_id=vi.img_id
LEFT JOIN view_img_dir AS vid ON vid.dir_id=vidr.dir_dir_parent_id
WHERE b.activate='1' AND pr.poller_id={}
```

When there is no icon, the `CONCAT` returns `NULL`, so the existing `res.value_is_null(10) ? "" : res.value_as_str(10)` already handles this case.

Two other small things (not blocking for map):

- `icon_image_alt`: for normal services this is `esi_icon_image_alt`, a free text set on the service itself, not something from the media. Here you use `view_img.img_comment`, which is the comment of the media. So it is not the same meaning for the same field. But `mod_bam` has nothing equivalent, so maybe `img_comment` is the best we can do.
- `icon_id`: you fill field 87, but the engine never sets it for normal services (`broker.cc` does not touch it). This is not a problem, I just note that BAs will be the only services with this field, so consumers cannot use it in a generic way.

jean-christophe81 (approved):

vuntz:
Not a code review, but from what I can tell, this should now work for map. Thanks!

PR #3699, opened 2026-09-11
fix(mbi): realign reporting collations on ETL runs

gfontaineCentreon:
## Description

MariaDB 11.5 and later make `utf8mb3_uca1400_ai_ci` the default collation of the utf8mb3 character set (see the `character_set_collations` system variable; 11.8 is the first LTS carrying the change). A table or a column created with a character set but no `COLLATE` clause therefore no longer shares the `utf8_general_ci` used by the rest of the MBI module, and comparing two different collations of the same character set raises `ERROR 1267 Illegal mix of collations`. That breaks the dimension joins on `hg_name` in `HostAvailability.pm` and `ServiceAvailability.pm`, and the availability aggregation stops.

The reporting tables are created from the structure of the central ones **only when they are missing** (`etl/import/main.pm::createTables` + `MySQLTables::createTable`), so a table that already exists keeps the collation it was created with and is never repaired. The import phase now realigns the ones that deviate, in the same action model as the table creation, so it happens on every ETL run.

Two deviations are handled, and both matter:

- a **column** collation, which is what those joins compare — needs the table to be rebuilt, so it is limited to the tables that actually deviate;
- a **table default** collation, which is what any column added to that table, or re-declared without a `COLLATE` clause, would inherit — a metadata-only change.

A table mixing several character sets, or one whose default collation was migrated out of the utf8mb3 family, is left alone entirely and reported instead: converting it would transcode its content or revert that migration.

The realignment stays out of the way of the run it travels with. Inspecting the schema cannot abort the run, and its statements are flagged `continue_on_error` so that one failing `ALTER` is reported and the next one still attempted, rather than stopping the scheduler and abandoning the dimension, event and perfdata phases. The tables left for a manual realignment are reported through the connector (gorgone log + progress stream), because this phase runs in the scheduler process, whose message collector is never drained.

The two transient tables that declared a character set without a collation are pinned as well, and the unit test schema is realigned on the module DDL — it carried the very declarations that cause the bug, so on the `mariadb:11.8` the unit tests run against, it was reproducing it.

**Fixes** [MON-208932](https://centreon.atlassian.net/browse/MON-208932)

Companion of centreon/centreon-modules#9264, which fixes the central side (module DDL + upgrade migration). Both are needed on a distributed architecture: that one repairs the monitoring databases, this one the reporting database.

## Type of change

## Target serie

- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

Validating on MariaDB 10.11 or 11.4, or on a platform whose reporting tables were created **after** the fix, proves nothing — the deviation only exists on tables created on MariaDB >= 11.5.

1. reporting server on MariaDB >= 11.5, with existing `mod_bi_*` tables. On an already-correct platform, reproduce the customer state by hand:
   `ALTER TABLE mod_bi_hostgroups MODIFY hg_name VARCHAR(255) CHARACTER SET utf8mb3 COLLATE utf8mb3_uca1400_ai_ci;`
2. run `centreonBIETL -d` (or the dimension + event phases) and observe the availability aggregation failing with `Illegal mix of collations`;
3. run the ETL again with this change: the import phase logs `[COLLATION] convert table [mod_bi_hostgroups] …`;
4. check that no `mod_bi_%` column of the reporting database is outside `utf8mb3_general_ci`, and that the availability aggregation completes and the reports generate.

Unit tests: `yath -L test ./gorgone/tests/unit/modules/centreon/mbi/` — the new `libs/bi/MySQLTables.t` runs against the CI MariaDB 11.8 and covers a deviating column, a deviating table default, a table mixing character sets (reported, not converted), the exclusion of the ETL transient tables, and idempotence. Verified locally that it fails when the fix is reverted, and that an unreachable database fails the file rather than passing it.

## Checklist

- [x] I have followed the **coding style guidelines** provided by Centreon
- [x] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [x] I have commented my code, especially **hard-to-understand areas** of the PR.
- [ ] I have made corresponding changes to the **documentation**.
- [x] I have **rebased** my development branch on the base branch (master, maintenance).

[MON-208932]: https://centreon.atlassian.net/browse/MON-208932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gfontaineCentreon:
## ✅ Automated Review — Passed

Reviewed with the Centreon automated review skill (code review with an error-handling focus, security, test coverage). No blocking issue remains; everything below was fixed in the branch before opening.

**Complexity:** high — the change emits DDL on a customer database on every ETL run — **Recommended reviewers:** 2, ideally one on MBI/ETL and one gorgone maintainer.

Findings that were applied:

- **The reports reached nobody.** `getCollationRealignStatements()` runs in the scheduler process, whose `Messages` collector is never drained — `getLogs()` is only called in `etlworkers/class.pm` — and `Messages.pm` maps `warning` to `I` without printing. The two branches that deliberately refuse to convert, i.e. exactly the tables that stay broken, pushed a string into an array discarded at the end of the run. The method now returns its warnings and the call site routes them to the gorgone log and to the progress stream, so they land in the `centreonBIETL` output.
- **One failing `ALTER` aborted the whole ETL run.** `sql()` has no per-statement guard, the connections are built with `die => 1`, the worker returns `FINISH_KO` and `action_centreonmbietllistener` moves the run to `STOP` — dimensions, event and perfdata all abandoned. For an opportunistic repair that is strictly worse than the previous behaviour, where only the availability reports failed. Statements can now be declared `continue_on_error`, and the schema inspection itself can no longer abort a run.
- **A reported table was not fully left alone.** Only converted tables were recorded as handled, so the second query still realigned the default collation of a table the first one had refused to convert — contradicting the message it emits. Reported tables are now recorded as handled too.
- **The test's failure mode was misleading.** `use tests::unit::lib::mockLogger;` was missing, so `centreon::common::logger` had no methods and any connection or SQL failure would have surfaced as `Can't locate object method "writeLogError"` instead of the real MariaDB error. Fixed, the local logger stub was replaced by the real `Messages` object, and the fixture statements now check their return.

Test coverage was also extended following the review: the mixed-character-set fixture now carries a deviating table default (which is what catches the third point above), and two fixtures pin the `mod_bi_tmp_%` / `%_tmp` exclusions.

> **Security review:** no findings. The interpolation of `information_schema` table names into `ALTER TABLE` was examined and rejected as a vector — no untrusted input reaches a `mod_bi_%` table name, `mysql_multi_statements` is not enabled on these connections, and the new code is more defensive than the code around it (`dumpTableStructure` has always interpolated without quoting). The conversion is utf8mb3 → utf8mb3, so collation-only: the guards make it impossible for a `utf8mb4` or binary column to be transcoded.

> **Test coverage:** `libs/bi/MySQLTables.t` runs against the CI `mariadb:11.8` and covers a deviating column, a deviating table default, a table mixing character sets, the exclusion of the ETL transient tables and idempotence — 14 assertions. Verified locally that it fails when the fix is reverted (`not ok 7`), and that an unreachable database fails the file instead of passing it. Not covered, and deliberately so: the call site in `createTables` (the action shape), and the `COLLATE` clauses of the two transient tables, whose real production names would collide with other tests on the shared database.

gorgone/gorgone/modules/centreon/mbi/etlworkers/import/main.pm
```diff
@@ -24,19 +24,41 @@ use strict;
...
+            next;
+        }
+
+        my $error;
+        try {
+            $connection->query({ query => $statement->[1] });
+        } catch {
+            $error = $_;
+        };
+        $etlwk->{messages}->writeLog('WARNING', $statement->[0] . ' failed: ' . $error) if (defined($error));
```
vuntz, line 61:
```suggestion
          try {
              $connection->query({ query => $statement->[1] });
          } catch {
              die $_ if (!$continueOnError);
              $etlwk->{messages}->writeLog('WARNING', $statement->[0] . ' failed: ' . $_);
          };
```

gorgone/gorgone/modules/centreon/mbi/etl/import/main.pm
```diff
@@ -56,6 +58,46 @@ sub createTables {
...
+            token => $etl->{run}->{token},
+            data => { messages => [ ['I', '[SCHEDULER][IMPORT] [COLLATION] ' . $warning ] ] }
+        );
+    }
+
+    if (scalar(@$collationStatements) > 0) {
+        push @{$etl->{run}->{schedule}->{import}->{actions}},
+            {
+                type => 1, db => 'centstorage', sql => $collationStatements,
+                continue_on_error => 1, actions => []
```
vuntz, line 97: We likely need to change class.pm to support continue_on_error.

vuntz (approved):

